PR #292, opened 2021-09-14
[WIP] Add multiple account supports

MainasuK:
- Add "AccountList" scene entry for switch between accounts
- Display notification count badge for different accounts in "AccountList" scene


MainasuK:
This feature works now. Merge and make i18n resource update.

Known issue:
The badge not implement yet.

Issue #267, opened 2021-08-05
[REQUEST]: remove character limit


Kibranoz:
Let's be honest. It really doesn't make sense. If I have a mastodon Post that's more than 500 characters I can just go into another app like Toot! And make my post. It's not based on any limits and not should there  be any limits to our speech.

MainasuK:
related #222

MainasuK:
The dynamic char limit works for servers which configure the char limits now. (https://github.com/mastodon/mastodon/pull/16485)

Issue #259, opened 2021-08-03
Add ability to click on follower list in Profile


OddHouseGames:
Add ability to click on followers on our profile (to bring parity to web version) and for community engagement so that you can follow users back if you wish.![image](https://user-images.githubusercontent.com/62911518/128099182-ef11bda3-cdfb-41f3-8697-fb5876f03552.jpeg)

MainasuK:
The feature is works in v1.2.0 (88) now.

Issue #224, opened 2021-07-31
It's very tiny on my iPad Mini 5


kyefox:
I'm not an iWhatzit dev, so this might be the expected behavior for an app not explicitly made for iPads.

![IMG_0048](https://user-images.githubusercontent.com/27457339/127724028-d8fe7080-9bd2-4b5e-ac54-982b36ab626a.PNG)

The expander will make it grow to fill most of the vertical width, but still leaves a lot of room on the sides.

Gargron:
This is normal for apps not made explicitly for the iPad

ShadowJonathan:
Are there any plans to accommodate an ipad interface?

Gargron:
In the future but it requires more design work

zeitschlag:
We support iPads now.

Issue #222, opened 2021-07-30
Support custom instance post character limits


cdmnky73:
Currently, the app sets a hard 500 character limit for posts, even on instances with higher limits

Should be able to get an instance's character limit and adjust the compose view accordingly

MainasuK:
related https://github.com/mastodon/mastodon/pull/16485

PR #303, opened 2021-11-03
Release v1.2.0

MainasuK:
Changes since v1.0.8 
- Add iPad supporting #224 
- Add follower/following list for user #259 
- Add multiple accounts supporting #220 
- Add dynamic compose character limit #222 #267 
- Update trends card UI
- Update notification mention segment UI
- Update i18n resources
- Update new global tint color
- Various bugfix


MainasuK:
Note:

The iPad implementation not release in v.1.2.0 (88) due to the delay for new UI. But it works in TestFlight v1.2.0 (86) if someone needs it.

PR #327, opened 2022-01-27
Update Timeline UI

MainasuK:
UI design:
- [x] Timeline & post detail 
- [x] Notification timeline
- [x] Profile timeline 
- [x] Profile header and profile fields tab
- [x] Post visibility indicator UI
- [x] Content warning UI 
- [ ] Post filter 

Also, the underlying persist logic now generate automatically via Sourcery. And
move from `Combine` redundant style concurrency logic to the new `async/await` style.


MainasuK:
Most UI/UX implemented. 

Merges and fix the "post filter" later.

PR #340, opened 2022-03-16
Handle onboarding authentication errors in /api/v1/instance

zacwest:
When `AUTHORIZED_FETCH` is enabled, the `/api/v1/instance` API response returns errors like:

```json
{"error":"This method requires an authenticated user"}
```

The server's still able to handle to handle the rest of the login flow, but the user experience of this error is an indefinite activity indicator that never goes away. The change here recovers from this error by filling in a minimal amount of discovery info for the instance which is enough to populate the server list and continue onboarding.

| Before | After |
| ------ | ----- |
|  |  |


MainasuK:
That's make sense. I'm not cover all the edge cases. It's looks pretty great.

Thank you.

Issue #384, opened 2022-04-26
[BUG] App shows wrong language


janlindblom:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The app is displaying Finnish labels instead of the expected Swedish or English ones. See attached image.
![AD614626-22A3-409F-8694-87F102B8DAB0](https://user-images.githubusercontent.com/154438/165302932-36e05808-417b-4e97-9a58-09900daf435c.png)

### Expected Behavior

App should have Swedish or English language labels, following the iOS settings and preference order.
![D78151C9-0AEE-45F6-B507-2AED9972D7D5](https://user-images.githubusercontent.com/154438/165303136-130ee23d-e6d3-4dc4-bf08-e05329e15334.png)

### Steps To Reproduce

Run the app in Finland with Finland as country/region and iOS language preferences set to Swedish, English, Finnish.

### Environment

```markdown
- Device: iPhone 11
- OS: iOS
- Version: 15.4
- Build: 19E241
```

himynameisjonas:
I have the same problem and I don't even have Finnish/Soumi listed in languages on my phone: 
![IMG_FFC598218508-1](https://user-images.githubusercontent.com/39345/165503886-a3be59d0-b0e7-447f-bcbd-8d02035afe0f.jpeg)

mikaelf:
I also noticed this and started a discussion: [Mastodon for iOS is in Finnish for Swedish people](https://github.com/mastodon/mastodon/discussions/18101).
I believe the issue is that the Finnish translation in the Mastodon for iOS app is in `Svenska (Finland)` (`Swedish (Finland)`). **Note** that there's a workaround to use English, that at least work on iPhones, see the linked discussion.

MainasuK:
<img width="792" alt="image" src="https://user-images.githubusercontent.com/7940186/165517981-1372cdff-0b40-4a63-8f0e-a5ba2244fdb2.png">

Yes, it's a bug. And the [Crowdin](https://crowdin.com/project/mastodon-for-ios) Swedish translation is still lacking. Please use the workaround and a build will release with unfinished Swedish i18n.

janlindblom:
Alright, then I’m not alone! I’ll see if I can spend a little time on Crowdin for the Swedish translation. Thanks!

janlindblom:
Ok, the Swedish translation is at 84% up from 19% now so should be a better experience!
[...]

PR #389, opened 2022-04-27
Fix sign up error prompt missing issue

MainasuK:
Ref: https://github.com/mastodon/mastodon/pull/15740

Refactor the sign up form and restore the error prompt label.


MainasuK:
Add Swedish i18n assets for #384

PR #395, opened 2022-04-28
fix: cursor jumping when entering text before non-ascii char in share extension

kimar:
# Rationale / Bug

When using the ShareActionExtension with a non-ascii char in the text input (e.g. an emoji) the cursor jumps to the end of the text when altering the text.

# How to reproduce

Open share action extension, enter any arbitrary text include any arbitrary emoji, move curse before emoji and type text

## Expected

Editing continues at cursor position

## Currently behavior

Cursor jumps to end of text

# Solution

Store current cursor position before and restore after text of `UIViewRepresentable` is being set.

| Before | After |
|---|---|
| Simulator Screen Recording - iPhone 13 Pro - 2022-04-28 at 21 04 18 mp4 | Simulator Screen Recording - iPhone 13 Pro - 2022-04-28 at 21 06 22 mp4 |

### DIFF
--- a/ShareActionExtension/Scene/View/StatusEditorView.swift
+++ b/ShareActionExtension/Scene/View/StatusEditorView.swift
@@ -54,6 +54,9 @@ public struct StatusEditorView: UIViewRepresentable {
     }
 
     public func updateUIView(_ textView: UITextView, context: Context) {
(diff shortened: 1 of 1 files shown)


Gargron (approved):

MainasuK:
That's looks great. 

Thank you!

PR #401, opened 2022-04-29
Proposal: use bundler to install Ruby Gems instead of installing them system-wide using root

kimar:
# Rationale

Currently Ruby Gems are installed globally which might lead to side-effects with other globally installed Gems. To mitigate this I'd propose to use Ruby Bundler and manage Gem versions using a `Gemfile.lock` for local usage.

# Pros

* Cleaner dependency management with pinned Gems
* Easier setup as installation of all required Gems will happen when calling a single command, in contrast to having to define all dependencies one-by-one when wanting to install them.
* Reduced maintenance as Gems are solely managed through Gemfile instead of having their existence spread throughout documentation and codebase.

# Cons

* Ruby Gem commands need to be prefixed with `bundle exec`


kimar:
@MainasuK you're right, `bundle` doesn't need to be installed using `sudo`, I've updated this. Apart from this, when doing a `bundle install` root permissions are required either.

PR #418, opened 2022-05-04
Improve tab bar icon vertical alignment

vollkorntomate:
I found that the tab bar icons feel a little misplaced, being not really centered. That's why I tried to improve their vertical alignment and move them to the center of the tab bar.
This only affected iPhone-sized devices, but not iPads.

This is how it was before:
tabbar-before

This is how it is now:
tabbar-after

Please let me know what you think.


MainasuK:
LGTM

Thanks

PR #466, opened 2022-07-13
Support flick up to pop the media

### LINKED ISSUE #464: [Feature] Swipe from bottom to the top should close current media
While watching the media files, e.g. images, the swipe from top of the screen to the bottom closes the media, but swipe from bottom to the top does absolutely nothing, which is annoying, since twitter and other software works both way. Can this behavior be implemented?

MainasuK:
resolve #464

Supports flick up to pop the media preview. And the original one from Apple could be found from here:
[PhotoTransitioning: : Using UIViewPropertyAnimator to create a fully interative and interruptible custom view controller transition](https://developer.apple.com/library/archive/samplecode/PhotoTransitioning/Introduction/Intro.html).


MainasuK:
Also, fix the profile banner without pop transition issue

PR #486, opened 2022-10-30
feat: handle paste event and insert images on the clipboard

### LINKED ISSUE #385: [BUG] Can't paste images from clipboard in New Post screen

### Current Behavior

https://user-images.githubusercontent.com/346590/165369630-d0cb82d7-9056-4d7f-b750-786e406a2173.mov

### Expected Behavior

https://user-images.githubusercontent.com/346590/165369587-a0fccc37-3f59-417c-9f12-dc8057a14a7b.mov

### Steps To Reproduce

See "Current Behaviour" section.

### Environment

```markdown
- Device: iPhone 12
- OS: 15.4.1
- Version: v1.3.1
- Build: 113
```

### Anything else?

_No response_

NanoSector:
Fixes #385 

The solution is quite ugly, since I force the event to propagate __everywhere__, however it seems somewhere down the line the `paste` event gets eaten/processed and it doesn't hit the ViewController anymore. Setting a debugger in the extension method in the if statement yields a UITableCellView instance as first 'capable' responder.

Suggestions are very welcome.


MainasuK:
We need to design a new viewModel to fit the new workflow. That needs the ability to transfer between states.

1. Idle (Wait for input)
2. Loading (Load the media from the library and create a thumbnail and file URL)
3. Uploading (Uploading…)
4. Failure (Bad input or upload failure with retry support)
5. Success (Ready for publishing)

zeitschlag:
Hey @NanoSector 

thank you very much for your contribution 🥳

I gave it a try and I'd suggest to postpone it to a later release until we implemented it, so that we can process images. When I pasted some text using your code, everything worked as expected, but when I pressed "paste" with an image, nothing happened.

When I used the current App Store-version in comparison, I could paste text, too, but using an image I didn't even see the paste-dialog. To me, this felt more intuitive, but ymmv.

So, how should we proceed? Would it be okay for you if we don't include it in the upcoming release? Otherwise we'd have to postpone the entire release. What do you think?

NanoSector:
Hi @zeitschlag ,

The original idea was to spawn an AttachmentService ad you can see by the commented code left in my last update. This worked & you would see an image be added as if you added an attachment with the paperclip button. You can see it in action here: https://social.nanosector.nl/@NanoSector/109258340896576173

However since the rewrite of the compose view in SwiftUI, last I checked attachments were completely non functional and I could not hook into this system anymore.

A long way of saying; the boilerplate code is there, it's just not hooked into anything anymore :)

MainasuK:
I appreciate your patience. I'm now ready to merge it. Please check the progress in #533

Issue #378, opened 2022-04-23
[BUG] VoiceOver improvements


TheQuinbox:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I really appreciate your efforts to make this app more accessible to VoiceOver users, and I already see improvement. However, it's not perfect. When on the details screen, you can't read the text of the current Toot. There are a few buttons with weird names, I think the Mastodon logo? At the top of the screen. And it would be extremely nice if we could have VoiceOver actions to do a lot of common things (e.g. like, reblog, etc). In addition, when I'm composing a toot, I can't read any of the text I typed. However, keep in mind I'm on a Mac, and I know Mastodon iOS isn't officially supported there. Not sure if that one happens on iOS.

### Expected Behavior

All the aforementioned issues should be addressed, making this more the most accessible Mastodon app on the market, and from an official source, no less.

### Steps To Reproduce

Try to navigate the app using the VoiceOver screen reader.

### Environment

```markdown
- Device: MacBook Air M1 (2020)
- OS: macOS 12.3.1
- Version: 1.3.1
- Build: 113
```

MainasuK:
Thank you for your feedback. 

There is a navigation bar title button in the home timeline. I add some label for differnet state of this button. It's will display in v1.4.0. And could you explain the how the VoiceOver action working? How can I trigger that action in Safari or others apps. I needs to learn it then try to fix it. 

In the compose and post detail page the app shoulds spells the content on iOS. I'm not try it on Mac. I'll take a notice of that.

TheQuinbox:
To access Voice Over Actions, you use a 1 finger swipe up and down. You can try it in, for example, Twitterrific from the Icon Factory.

ahundt:
Hi since this looks like an umbrella voiceover bug report I’m reporting the following issue encountered by a VoiceOver user. 

> I’m #Deafblind and using #VoiceOver, the iOS screenreader #blind people use. Mastodon’s button labeled Compose doesn’t work, but if I swipe right three times and double tap a button labeled “square.and chart,” then the publishing field appears! There’s a lot of visual content that’s not accessible here, too. What kind of world have I just entered?! #Accessibility #a11y

https://mastodon.social/@HabenGirma/109259674417438870

Bracciata:
Can this be merged with #309

j-f1:
An update on this issue:

- [x] When on the details screen, you can't read the text of the current Toot. _Fixed in #491!_
[...]

PR #488, opened 2022-10-31
Fix accessibility for the Compose button

j-f1:
Ref: https://github.com/mastodon/mastodon-ios/issues/378#issuecomment-1296377899, https://mastodon.social/@HabenGirma/109259674417438870

- hide the Compose button from VoiceOver
- allow tapping on the invisible Compose tab instead
- block switching to the Compose tab, open the modal instead.


MainasuK:
LGTM. I do not have a screen reader. And this looks should fix the button access issue. 

This patch will be a great help. Thank you for your contribution.

Issue #328, opened 2022-01-29
app is not accessible.


paxcoder:
I can't interact with toots using voiceover

MainasuK:
Hi @paxcoder,

I'm sorry the interactions are not fully handled. The app only carry with basic toot reading a11y supports now. Recently, I've figure out how to response to the VoiceOver properly. So the most a11y issues should be fixed in the next version.

TheQuinbox:
Ah, beat  me to it. Was just about to create this issue.

Most everything is accessible, it's just we can't do anything with toots because there are no VoiceOver actions, no buttons, and we can't click on the toot to bring it up in a different screen, even.

paxcoder:
but it sucks.
we don't have any accessible app that is not abandonware.
it's why we need to use apps that was updated 3 years ago, or in the best case 9 months ago.

paxcoder:
bump up, several months later, we are dumped. this app is abandonned too?

MainasuK:
I know it's been a long time since I updated. But I don't have time to do it at the moment. I will try to get some updates after some busy with other projects.

And I wonder why VoiceOver not working? I testing the home timeline and it's should read the post content and doulbe tap will redirect to the converstaion page which allows you interacte with that. And I'm sorry for the missing VoiceOver actions. I cannot give any promise about update time. But that's will be fixed once I'm free from other projects.

edhowland:
I am using version 1.5 of the IOS app on iPhone 12 mini, IOS version 16.3
The only noticable  VoiceOver problem I am currently having is the scrolling.

Turn on VoiceOver.
Go to almost any list or thread.
When you swipe right to see the next post and you reach the bottom of the curren page on screen it leaves the thread and does not automatically scroll down a page. 
It leaves the threads and jumps down to the tab bar.
You can work around this by going over to the scroll bar and swiping down one time and then going back over to thethread list.
You start at the top on page 2 and go to the bottom of page 2 where the same behaviour happens again.

Let me know if I can provide any other information.

PR #491, opened 2022-10-31
Improve accessibility of the thread screen

j-f1:
- combine the avatar, display name, username, timestamp, menu, and eye button into a single accessibility element
- set an appropriate label for that element that includes the relevant information
- attach custom accessibility actions for the actions formerly in the “…” menu
- attach an enable/disable content warning when the corresponding button is visible
- improve labelling of the reblog & favorite buttons
- remove the status metric view from the accessibility hierarchy
  - read the full date after the relative timestamp in the author view
  - add “show reblogs” and “show favorites” as custom actions on the reblog & favorite buttons

I moved the author section into its own subview so the accessibility properties can be directly controlled. Let me know if I should undo some of the changes I made to move code into that subview :)


j-f1:
One issue I’m seeing (and I’d guess that it’s the underlying cause of #328) is that I’m unable to go past the “Post from [name]” header of a thread in VoiceOver, but tapping on the header seemingly takes me to a identical-sounding but different title that I can navigate past. Covering it up for now by navigating to the table view directly when showing the thread, but if you cancel a reply you’ll still end up in the bad state.

j-f1:
Should this one also be retargeted to `develop`?

MainasuK:
> Should this one also be retargeted to `develop`?
Yes. We use git-flow and you can check [this](https://github.com/mastodon/mastodon-ios/blob/main/Documentation/CONTRIBUTING.md). 


> One issue I’m seeing (and I’d guess that it’s the underlying cause of https://github.com/mastodon/mastodon-ios/issues/328) is that I’m unable…

I need more context to understand that problem. The app currently read the whold post in the home timeline. The reactions (reply, repost, like…) only available after tap to enter the detail page. This behaivor is some mimics from Twitter app. (I wonder waht is the best practical)

- The cell configuration using `Core Data` entity's publisher to bind the property to UI. That's maybe not fit the Voice-Over to get the correct value.
- There is a VoiceOver menu or something missing for reactions.

MainasuK:
> Covering it up for now by navigating to the table view directly when showing the thread, but if you cancel a reply you’ll still end up in the bad state.

I have to admit the lack of knowledge about VoiceOver. And this changes should make sense.

MainasuK:
The code conflict is due to the new menu entry for bookmarks. That's too late and I'll handle this tomorrow.

Also, thank you for your contributions.

j-f1:
> I need more context to understand that problem. The app currently read the whold post in the home timeline. The reactions (reply, repost, like…) only available after tap to enter the detail page. This behaivor is some mimics from Twitter app. (I wonder waht is the best practical)

The issue here is that when you tap on a post in the timeline you're taken to the detail view as expected but your cursor starts off on the title in the navigation bar, and there's no way to navigate to the post actions for some reason. This might well be a bug in iOS since I am struggling to identify the conflicting element using Accessibility Inspector.

MainasuK:
LGTM. Thank you for your works.

Issue #339, opened 2022-03-15
Custom App Icons


Suzie97:
## Proposal
Add some custom app icons for the app. Just for fun.

### GitHub for iOS
![Untitled](https://user-images.githubusercontent.com/68198116/158335414-6b787ccc-0062-4744-8fa0-42dbb5ea0c3d.png)

I've made some custom icons for fun. But I would be happy to contribute them to the project if needed.
<img width="532" alt="Screenshot 2022-03-15 at 13 51 31" src="https://user-images.githubusercontent.com/68198116/158335632-1e3a6d9c-96b7-403e-9e1b-8cab641878d3.png">

Gargron:
We will consider adding customization options in the future, but we are still building out core features.

patrick91:
@Gargron would you be willing to accept a PR from the community for this feature? ☺️

AugustinMauroy:
Can i contribute on this pr ?

PR #495, opened 2022-11-01
Add an alternate app icon for debug builds

j-f1:
This makes it easier to tell the debug and production apps apart. I did a basic hue rotation and am totally open to different colors or some other alteration, I just don’t want to have to wonder which app I’m using.

IMG_7862

Icon created using…

```sh
find *.png -exec magick \{\} -modulate 100,100,185 ../AppIcon-Debug.appiconset/\{\} \; 
```


Suzie97:
image

I made this icon and i think this one will be a better icon for the debug version. I can provide the required files if needed.

j-f1:
Ooh I like that one a lot better!

Gargron:
That looks pretty cool, thanks!

MainasuK:
https://github.com/mastodon/mastodon-ios/issues/339

Ohhh.
Maybe we really need to add the custom App Icon features.
So users could choose their favorite one. @Gargron 

What I need is a design about the setting entry and how to display icons.  

P.S.
And the app needs the opaque (no alpha) png or jpeg icon image in 1024 x 1024.

j-f1:
Closing this for now since the changes I made were fairly simple and I’m no longer blocked (I just keep a third-party app around for notification purposes)

PR #496, opened 2022-11-01
Fix typos in Setup.md

jordanekay:


MainasuK:
Thanks a lot. 

I change the merge target to develop. So the `CocoaPods-Keys` part is removed. And now the repo default branch is set to `develop` to prevent the confusion about the git-flow we using.

PR #500, opened 2022-11-03
Add support for scrolling the discovery tab to the top/first tab

j-f1:
When tapping on the 🔍 icon in the tab bar while it is selected, the current scroll view will be scrolled to the top. If it’s already at the top, the “Posts” tab will be selected. This matches how all the other tabs scroll to the top when their tab item is tapped again.

I grabbed the “scroll to first tab” behavior from Twitter but am also happy to remove it.


MainasuK:
I'm not sure scroll to post is necessary. But we can make it fly first. 

Again. Thank you for your contributions!

PR #502, opened 2022-11-03
Add accessibility actions for links/mentions/hashtags in posts

j-f1:
- Applies to either the post text in the expanded style, or to the entire post view in other styles.
- Removes the duplicate accessibility label which caused post text to be read twice
- actions are not available when a content warning is applied to the post and the post content is not currently visible


j-f1:
In terms of wording, here’s what I’m seeing on centralized social media apps:
- Twitter uses basically the same language I use here (“@[user],” “#[hashtag]” and “Link: [link]”)
- LinkedIn uses the term “View [user],” “Search #hashtag,” and “Open [link]”
- Facebook puts the links/mentions/hashtags as separate accessibility elements after the post content, requiring users to swipe through them (but also meaning you don’t need to swipe up/down to access them)
- YouTube community posts put the links as actions without any labelling, and video descriptions use the Facebook style

MastodonSDK/Sources/MastodonUI/Extension/MetaEntity+Accessibility.swift
```diff
@@ -0,0 +1,26 @@
...
+    var accessibilityCustomActionLabel: String? {
+        switch meta {
+        case .url(_, trimmed: _, url: let url, userInfo: _):
+            return "Link: \(url)"
+        case .hashtag(_, hashtag: let hashtag, userInfo: _):
+            return "Hashtag \(hashtag)"
+        case .mention(_, mention: let mention, userInfo: _):
+            return "Show Profile: @\(mention)"
+        case .email(let email, userInfo: _):
+            return "Email address: \(email)"
```
zeitschlag, line 20: As far as I understand, only people speaking English would benefit from it right now. Could you sprinkle in some localization, please? 🙂

j-f1: That’s definitely a major oversight, thanks for pointing it out!

zeitschlag: Adding a `//TODO:` doesn't fix it :D

j-f1: Is there something else I can do now? I think when @MainasuK asked about localization earlier they said the `L10n` stuff is auto-generated so it would have to be subbed in later once the PR is merged and Crowdin runs. If there’s more I can do now I’m happy to do so tho! (I already added the corresponding keys to `app.json`)

zeitschlag: Ah, great, I see 👍  What I wanted to suggest is adding them to the EN `Localizable.strings`, run `swiftgen` once by hand and use the `L10n.keys` so that everything is prepared for the translated strings. Do you know what I mean?

j-f1: Looking into it now and it seems like the top-level `a11y` key here conflicts with the one defined in the stringsdict resulting in an invalid redeclaration error (https://github.com/SwiftGen/SwiftGen/issues/1017, but a different cause). I don’t want to manually update the generated file, do you have thoughts on a path forward?

j-f1: Figured out a workaround

MastodonSDK/Sources/MastodonUI/Extension/MetaEntity+Accessibility.swift
```diff
@@ -0,0 +1,27 @@
...
+    var accessibilityCustomActionLabel: String? {
+        switch meta {
+        case .url(_, trimmed: _, url: let url, userInfo: _):
+            return L10n.Common.Controls.Status.MetaEntity.url(url)
+        case .hashtag(_, hashtag: let hashtag, userInfo: _):
+            return L10n.Common.Controls.Status.MetaEntity.hashtag(hashtag)
+        case .mention(_, mention: let mention, userInfo: _):
+            return L10n.Common.Controls.Status.MetaEntity.mention(mention)
+        case .email(let email, userInfo: _):
+            return L10n.Common.Controls.Status.MetaEntity.email(email)
```
zeitschlag, line 21: Great, that's what I meant! 👍

Localization/StringsConvertor/Sources/StringsConvertor/Parser.swift
```diff
@@ -43,7 +43,12 @@ extension Parser {
                             .map {
                                 switch keyStyle {
                                 case .infoPlist:        return $0
-                                case .swiftgen:         return $0.capitalized
+                                case .swiftgen:
+                                    if $0 == "a11y" {
+                                        return "A11y"
+                                    } else {
+                                        return $0.capitalized
+                                    }
```
zeitschlag, line 51: I'm a bit confused. You changed the keys in `app.json`, which means: This isn't necessary anymore, right? Or is there anything I don't see? Sorry for being a bit nitpicky 🙃

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
```diff
@@ -315,7 +315,6 @@ extension StatusView.ViewModel {
                 statusView.contentMetaText.configure(
                     content: content
                 )
-                statusView.contentMetaText.textView.accessibilityLabel = content.string
```
zeitschlag, line 318: What's the reason for this? The `accessibilityCustomActions` aren't a replacement for `accessibilityLabel` from my understanding. Or am I wrong?

zeitschlag: Ah, sorry, nevermind, that's what you meant with

> Removes the duplicate accessibility label which caused post text to be read twice

right?

j-f1: That’s correct.

Localization/StringsConvertor/Sources/StringsConvertor/Parser.swift
j-f1: Yep, reverted!

Issue #365, opened 2022-04-07
Add "Hide reblogs" dropdown option on profiles


Gargron:
When you are following a user, you can control if you see their reblogs on your home feed.

PR #505, opened 2022-11-04
Show/Hide Reblogs (#365)

zeitschlag:
https://user-images.githubusercontent.com/2580019/200420833-943723ab-fb72-48b8-8415-290d0023b9bc.mov

### DIFF
--- a/Localization/app.json
+++ b/Localization/app.json
@@ -460,11 +460,11 @@
                 },
                 "confirm_show_reblogs": {
                     "title": "Show Reblogs",
-                    "message": "Confirm to show reblogs
+                    "message": "Confirm to show reblogs"
                 },
                 "confirm_hide_reblogs": {
                     "title": "Hide Reblogs",
-                    "message": "Confirm to hide reblogs
+                    "message": "Confirm to hide reblogs"
                 },
             },
             "accessibility": {
--- a/Localization/app.json
+++ b/Localization/app.json
@@ -180,7 +180,9 @@
                 "unmute": "Unmute",
                 "unmute_user": "Unmute %s",
                 "muted": "Muted",
-                "edit_info": "Edit Info"
+                "edit_info": "Edit Info",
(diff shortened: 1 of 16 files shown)


Localization/app.json
```diff
@@ -455,7 +457,15 @@
                 "confirm_unblock_user": {
                     "title": "Unblock Account",
                     "message": "Confirm to unblock %s"
-                }
+                },
+                "confirm_show_reblogs": {
+                    "title": "Show Reblogs",
+                    "message": "Confirm to show reblogs
```
MainasuK, line 463: It looks like the suffix `"` is missing.

Localization/app.json
```diff
@@ -455,7 +457,15 @@
...
+                },
+                "confirm_hide_reblogs": {
+                    "title": "Hide Reblogs",
+                    "message": "Confirm to hide reblogs
```
MainasuK, line 467: It looks like the suffix `"` is missing.

MastodonSDK/Sources/MastodonUI/ViewModel/RelationshipViewModel.swift
```diff
@@ -27,6 +27,7 @@ public enum RelationshipAction: Int, CaseIterable {
     case edit
     case editing
     case updating
+    case showReblogs
```
MainasuK, line 30: Maybe the `Reblogs` not needs here. Or maybe insert as a low priority relationship. This view model will drive the `ProfileRelationshipActionButton` to display different states.

The priority drive by implicitly enum Int value. So move `showReblogs` before ``

For example, 
If one person is blocking by "myself" and is muting by "myself". Only the high priority state (here is `blocking`) display. Once we unblock the person. The lower priority state display. 

So move the `showReblogs` case before `isMyself` also could be a fix.

MainasuK (commented):
Please check the profile and make sure that button works due to the relationship logic changed.

PR #516, opened 2022-11-06
Mark sidebar cells as buttons

j-f1:
@<span>Marco@</span>tech.lgbt [pointed out](https://tech.lgbt/@Marco/109296496757761928) that the items in the sidebar are not properly identified as buttons, but said they’re otherwise fine. This PR adds the missing trait.

> [the compose button on iPad] sounds different than the iPhone one with VoiceOver, so may just need to receive the button trait.
> [me: can you still activate the button?]
> I can activate them just fine. So giving them the proper traits should be enough.

Note: I’m unable to test this because I don’t have an iPad, running Mastodon in the iPad simulator results in crashes, and VoiceOver doesn’t work directly in the simulator.

### DIFF
--- a/Mastodon/Scene/Root/Sidebar/SidebarViewModel.swift
+++ b/Mastodon/Scene/Root/Sidebar/SidebarViewModel.swift
@@ -141,6 +141,7 @@ extension SidebarViewModel {
             cell.setNeedsUpdateConfiguration()
             cell.isAccessibilityElement = true
             cell.accessibilityLabel = item.title
+            cell.accessibilityTraits.insert(.button)
         }
         
         // header


zeitschlag:
Hi @j-f1,

thank you for this PR 🥳 

What do you mean by this:

> running Mastodon in the iPad simulator results in crashes

I just fired the Simulator (and the app) up and it worked fine. Can you reproduce that (and could you please open an issue for that, if so)?

j-f1:
Sure thing, opened #522

zeitschlag:
@j-f1 FYI: I just merged it, although the PR added only the button-trait for the `compose`-button, but not the other cells 🙂 I gonna take care of that.

Again: Thank you very much for the PR 👍

PR #520, opened 2022-11-07
Improve accessibility of the account switcher

j-f1:
- mark the DragIndicatorView as a button and use `accessibilityActivate` instead of a tap gesture recognizer
- remove the avatar buttons from the VoiceOver flow (esp. since they’re disabled)

Note: I tried to update the translation strings for `a11y.plural.count.unread.notification` which currently say “0 unread notification.” Did I do it right / does anything else need to happen to update the English translation?


MainasuK:
Please check [here](https://github.com/mastodon/mastodon-ios/tree/develop/Localization) about the i18n.

You can append the new string into the JSON file. But due to the i18n workflow delay. The L10n will be ready until the i18n branch is merged. Maybe the workflow could be improve so that `L10n` could be always available for new string.

j-f1:
I looked in that file but it doesn’t seem to have the particular string I was editing. However, I’ll try that approach in my other branches!

j-f1:
Reverting f9daeea4d3ad66a5387493b67682c4e27396d2bf because the custom action doesn’t seem to be showing up for me on the latest iOS 16.2 beta. I could’ve sworn it worked before, not sure if it’s a beta bug or I just imagined it.

MastodonSDK/Sources/MastodonLocalization/Resources/en.lproj/Localizable.strings
```diff
@@ -154,7 +154,7 @@ Your profile looks like this to them.";
 "Common.Controls.Timeline.Timestamp.Now" = "Now";
 "Scene.AccountList.AddAccount" = "Add Account";
 "Scene.AccountList.DismissAccountSwitcher" = "Dismiss Account Switcher";
-"Scene.AccountList.TabBarHint" = "Current selected profile: %@. Double tap then hold to show account switcher";
+"Scene.AccountList.SwitchAccounts" = "Switch Accounts";
```
MainasuK, line 157: That will not take effect. Suggest rollback changes.

```suggestion
"Scene.AccountList.TabBarHint" = "Current selected profile: %@. Double tap then hold to show account switcher";
```

MainasuK:
Hello, @j-f1. The i18n workflow now update. You can add the new string in the template JSON / stringsdict file under `./Localization`. Then just run `update_localization.sh` to create the new `L10n` and use that directly.

I see some changes in `./Localization/StringsConvertor/en.lproj`. And that's will not take effect and will be overridden by Crowdin sync. 

And this PR is LGTM and ready for merge. Will you want to update some strings or just let me merge it?

MastodonSDK/Sources/MastodonLocalization/Resources/en.lproj/Localizable.strings
j-f1: Done!

PR #521, opened 2022-11-07
Improve accessibility of the compose view

j-f1:
- add missing labels to the toolbar
- make the avatar view easier to understand + navigate past


MainasuK:
LGTM. I think the new compose scene is ready now.

Thank you!

PR #552, opened 2022-11-09
Improve accessibility of the notifications tab

### LINKED ISSUE #541: [BUG] Accessibility/VoiceOver: Not intractable text on the notifications overview will be called out as buttons

### Current Behavior

- Go to the Notifications screen
- Turn on VoiceOver
- Swipe right to go to the content

**Issue 1**: Small "dot"will be announced and called out as a button

IMG_9044

**Issue 2**: Timestamp will be called out as a button

IMG_9045

### Expected Behavior

Issue 1:
- The small "dot" shouldn't be announced via VoiceOver at all

Issue 2:
- The timestamp should be announced, but not called out as a button

### Steps To Reproduce

See above

### Environment

```markdown
- Device: iPhone 12 mini
- OS: iOS 16.1
- Version: 1.4.5
- Build: 144
```

### Anything else?

_No response_

j-f1:
Fixes #541. Now, each notification appears as a single element to VoiceOver. Appropriate secondary actions are added (all items from “…” menu, confirm/delete for follow requests, “show user profile” (for non-follow/follow request notifications)).

Note that it’s not yet possible to reply/reboost/favorite a post from the notifications screen without activating it to open its thread. (I’m not sure if this limitation should be fixed or if it makes more sense to keep the set of actions folks have to scroll through fairly small). On that note, one thing I’ve been considering is how several other social media apps condense a bunch of actions into one that opens an action sheet / list of buttons. Would it make sense to do that here? (For example, Twitter puts the actions from the “…” menu under a “More Actions” action, while leaving like/retweet/reply as top level actions).

I’ve also adjusted the descriptions of reblog/reply posts. They now read as “User B replied to User A” and “User A reblogged User B.” This fits better with the grammar in English, but do you think there would need to be new localized strings to handle grammatical differences between languages?


MastodonSDK/Sources/MastodonUI/View/Content/NotificationView+ViewModel.swift
```diff
@@ -125,6 +125,76 @@ extension NotificationView.ViewModel {
...
+
+        Publishers.CombineLatest(
+            $authorAvatarImage,
+            $type
+        )
+        .sink { avatarImage, type in
+            var actions = [UIAccessibilityCustomAction]()
+
+            // these notifications can be directly actioned to view the profile
+            if type != .follow, type != .followRequest {
```
kimar, line 158: Can't we also append this action to a `follow` event? Without Voice Over I'm able to open up the profile from a follow-notification, with Voice Over enabled I'm not anymore.

kimar:
Hi @j-f1 thank you first of all for this PR, I think it greatly improves the UX on the Notifications Screen, code-wise it looks fine to me, just two things I noticed while testing:

* I'm not able to open the user profile when coming from a follow notification
* The action menu is discoverable through the Rotor but when letting voice over read the on-screen items I'm not getting informed about the existence of the actions menu

I would definitely like to get the improvements merged, maybe need a concept for how to handle the cases you mentioned in the description.

MastodonSDK/Sources/MastodonUI/View/Content/NotificationView+ViewModel.swift
j-f1: You should be able to use the default “activate” action to open the profile for these types of notifications, whereas for other notifications that action will open the corresponding post instead.

j-f1:
> The action menu is discoverable through the Rotor but when letting voice over read the on-screen items I'm not getting informed about the existence of the actions menu

On my iPhone 13 (running the latest iOS 16.2 beta), VoiceOver reads “Actions available” at the end of the label for each notification. I saw some reports on Stack Overflow that some iOS versions may not read that label for elements which have the same set of actions as a previously read element, though.

j-f1:
In terms of the cases mentioned in the description, my thoughts are:

- one option would be to simply add the additional actions (“reply,” “boost,” “favorite”) to the set of custom actions
- Another would be to additionally remove the options from the “more” / “…” menu and add a new “More Actions” action that opens an action sheet with the options from the More menu to reduce clutter in the set of custom actions (especially as we add additional actions for the links/mentions). We have to have an action sheet since `UIMenu`s can’t be programmatically opened.

MastodonSDK/Sources/MastodonUI/View/Content/NotificationView+ViewModel.swift
j-f1: That’s very weird, on iOS 16.2 the action works fine. Maybe it has something to do with focus or something? I don’t have a person following me at the top of my notification feed so I guess perhaps there could be an issue with the first one or something? If you’re able to figure out in more detail what’s going on that would be awesome :)

PS: You can turn on the “Caption Panel” (at the bottom of VoiceOver settings) to get a small panel at the bottom of your screen that lets you visually read what VoiceOver is saying, which can be useful for debugging accessibility issues sometimes)

DK1DA: @j-f1 I managed to trigger the default open action on the item now, looks like I released my finger to quickly. That being said, if you resolve the conflicts we could merge this PR.

kimar:
Tank you very much @j-f1 🙌

PR #559, opened 2022-11-10
Add SwiftLint and SwiftFormat but with no settings

dalemyers:
The purpose of this change is to introduce the tools and show that they have no effect with all rules disabled. The rules that are chosen can be very contentious in any developer community, so this PR deliberately avoids having any turned on and instead just enables the tools so that rules can be handled in later PRs.


dalemyers:
Here's an example PR which could come later that has a _ton_ of change: https://github.com/dalemyers/mastodon-ios/pull/1/files

Yonodactyl:
I know your example just removes the duplicates, but seeing 562 Files changed in a PR gives me anxiety.

Like the idea of SwiftLint though - 👍🏼

dalemyers-flipdish:
Agreed. That's why I wanted to add the tools with no rules and then add them one at a time. 500+ files is a nightmare PR, but since its just a very small change, in this case ordering imports, then each change only takes a second or two to review, making this a 10-20 minute PR. 

It's not ideal, but it's the best option going forward I feel.

dalemyers:
Agreed. That's why I wanted to add the tools with no rules and then add them one at a time. 500+ files is a nightmare PR, but since its just a very small change, in this case ordering imports, then each change only takes a second or two to review, making this a 10-20 minute PR.

It's not ideal, but it's the best option going forward I feel.

Yonodactyl:
I'm with you on this, @dalemyers 
Small improvements make big improvements

kimar:
@dalemyers thank you for the PR, I like SwiftLint as well but as you already noted it would be a huge undertaking, and currently there is a lot of flux in this Repo, also only having SwiftLint without any rules enabled would add maintenance cost without bringing real benefit. I'm going to close this PR for now and we're going to internally talk about this topic and decide on which SwiftLint rules we're going to want to use.

Issue #312, opened 2021-12-05
Verification label on user profiles URLs


jbalint-de:
This is an obvious suggestion and feature request to introduce verification marks for verified links in a users profile. 

Thanks for your work. 🙂

edent:
User need identified at 

https://mastodon.social/@b0o@mstdn.io/109304553566634986

j-f1:
Fixed by #564.

PR #564, opened 2022-11-11
Visual indication of validated URL

ReactiveRaven:
Adds a visual indication if fields in a profile have been validated as associated with the account

The colours are based on those used in the web-app.

| Light theme | Dark theme | Black theme |
|-----|-----|-----|
|  |  |  |

On tapping the checkmark inside a validated field, a popover shows the time that the server verified the url

RE: #312


j-f1:
I don’t know if it’s possible but do you think the link text should also be colored green?

Mastodon/Diffiable/Profile/ProfileFieldSection.swift
```diff
@@ -57,7 +58,17 @@ extension ProfileFieldSection {
...
+            if (field.verifiedAt.value != nil) {
+                backgroundConfiguration.backgroundColor = Asset.Scene.Profile.About.bioAboutFieldValidatedBackground.color
+            }
             cell.backgroundConfiguration = backgroundConfiguration
+            
+            // set checkmark
+            cell.checkmark.isHidden = true
+            if let verifiedAt = field.verifiedAt.value {
+                cell.checkmark.isHidden = false
+                cell.checkmark.accessibilityLabel = "Ownership of this link was checked on \(verifiedAt)" // TODO: I18N / L10N
```
j-f1, line 70: It would be cool to surface this for folks who don’t use accessibility tools

ReactiveRaven: I agree, but I'm not sure how best to display it. A few other mastodon clients I tried only show it in voiceover, and the web-app relies on the 'title' attribute which doesn't work with touch input.

Do you have any suggestions?

ReactiveRaven: I've gone with a simple UIAlertController when you tap the checkmark, for now:

ReactiveRaven:
> I don’t know if it’s possible but do you think the link text should also be colored green?

The link is now green too, @j-f1

Mastodon/Diffiable/Profile/ProfileFieldSection.swift
ReactiveRaven: I realised I'd been using the wrong term; I've updated all variants of 'Validated' to 'Verified'.

j-f1: Another option would be to use the edit menu APIs (for example, try tapping a timestamp in the GitHub iOS app)

ReactiveRaven: OK, that has been re-implemented with UIMenuController and UIEditMenuInteraction depending on the OS version. It does mean it can be contained in the ProfileFieldCollectionViewCell, which is nice :)

I've had to shorten the text so it doesn't get truncated, but it does expose the time of verification now.

Gargron:
From a design perspective, this looks exactly like what we would've gone with. I'm not qualified to comment on the code but hopefully we can merge this soon.

Localization/app.json
```diff
@@ -438,6 +438,10 @@
                 "placeholder": {
                     "label": "Label",
                     "content": "Content"
+                },
+                "verified": {
+                    "short": "Verified at %s",
```
j-f1, line 443: Slight grammar tweak (since it includes the date):

```suggestion
                    "short": "Verified on %s",
```

PR #565, opened 2022-11-11
Update navigation bar appearance in SearchViewController

kylebshr:
Since discover has an opaque paging tab bar underneath the navigation bar, it looks pretty weird to have a translucent navigation bar and a double separator. This PR updates the navigation bar appearance of the search view controller so that it blends with the paging tabs.

I think there's a good argument to be made that the paging tab background should be updated to match the navigation bar (especially on iPad) - however, I couldn't configure a `UINavigationBar` nor a `UIVisualEffectView` to match the navigation bar styling. Since we can't do that, this seems like a reasonable solution for now.

| | Before | After |
|-|-|-|
| iPhone (Light) | IMG_0608 | IMG_0607 |
| iPhone (Dark) | Simulator Screen Shot - iPhone 14 Pro - 2022-11-11 at 15 47 28 | Simulator Screen Shot - iPhone 14 Pro - 2022-11-11 at 15 07 42 |
| iPad (Light) | Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2022-11-11 at 15 44 04 | Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2022-11-11 at 15 42 38 |
| iPad (Dark) | Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2022-11-11 at 15 44 12 | Simulator Screen Shot - iPad Pro (11-inch) (4th generation) - 2022-11-11 at 15 42 20 |


kylebshr:
Closing until I've fixed a couple issues

kylebshr:
Should be good now!

Mastodon/Scene/Search/Search/SearchViewController.swift
```diff
@@ -73,19 +73,19 @@ extension SearchViewController {
...
         super.viewDidLoad()
 
-        setupBackgroundColor(theme: ThemeService.shared.currentTheme.value)
+        setupAppearance(theme: ThemeService.shared.currentTheme.value)
         ThemeService.shared.currentTheme
             .receive(on: DispatchQueue.main)
             .sink { [weak self] theme in
                 guard let self = self else { return }
-                self.setupBackgroundColor(theme: theme)
+                self.setupAppearance(theme: theme)
```
Yonodactyl, line 81: Is there a benefit to calling this function twice, or would calling it in the Sink be enough?

kylebshr: I don't think the sink is called unless the theme changes; the initial call is needed to set things up with the current theme. (That's the precedence I’ve seen in the code base, I didn't actually check though)

Yonodactyl: That makes sense! I'm still familiarizing myself with Combine and the concept of disposable bags, so excuse my ignorance!

kimar (approved):
LGTM 👍  🎉 Thank you @kylebshr I'm going to merge this.

Issue #174, opened 2021-06-23
Add pull to refresh to thread view


Gargron:
Suggestion from TestFlight feedback

OddHouseGames:
Would be super useful to be able to refresh to see the reply I posted.

Currently when I post a reply, I have to back out of the conversation and enter it again to be able to see that my response posted correctly.

Or could do what Twitter does and show banner;![image](https://user-images.githubusercontent.com/62911518/135592910-36248d68-e830-4627-a864-b14b82504aeb.jpeg)

PR #583, opened 2022-11-14
Feature: Display just published toots in the thread immediately

leostera:
Hello folks! Thanks for the app ❤️‍🔥 

I'm a recent migrant from twitter and noticed in one of my recent tootnamis that I had to recur to this tiny dance:

1. toot something
2. tap reply
3. toot again
4. go out and back in to the toot in step 1 to see my toot from step 3
5. repeat steps from step 2 

And this adds a ton of awkwardness when doing stream-of-consciousness writing. If this isn't a valid use-case that's fine, but if it should be, I'd love to help make it happen.

In this branch I've been toying around with the codebase, and I've managed to get a flow that looks like this: 

https://user-images.githubusercontent.com/854222/201650281-61ce99fe-571d-4962-a08a-32856d27875f.mov

Would this be a welcome addition? If so I'd be happy to get some pointers on how to actually get the data flowing in the right way, and fix some other bugs I've seen with it (like a reply not moving the current thread root, so if you tap to reply on your reply, it is hidden!).

In any case, it was fun doing this live and toot about it as I went along: https://mas.to/@leostera/109340850992406029

### DIFF
--- a/Mastodon/Scene/Compose/ComposeViewModel+PublishState.swift
+++ b/Mastodon/Scene/Compose/ComposeViewModel+PublishState.swift
@@ -138,6 +138,7 @@ extension ComposeViewModel.PublishState {
                     }
                 } receiveValue: { response in
                     os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s: status %s published: %s", ((#file as NSString).lastPathComponent), #line, #function, response.value.id, response.value.uri)
+                    self.viewModel?.publishedStatus = response.value
                 }
         }
     }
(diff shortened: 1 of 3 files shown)


MainasuK:
That looks useful. First, thank you for your contribution.

And there are some cases that maybe needs an explain. The current iOS app display conversation replies in tree mode with max(2) tier. Tier 3 will be hidden and only the reply count number displayed. So a typically conversation looks like this:

```
Root  (A)
|-tier 1 (someone p1, reply to A, as A')
|-tier 2 (someone p2, reply to A', as A'')
|-tier 1 (someone p3, reply to A, as B)
|-tier 1 (someone p4, reply to A, as C)
|
…
```

1. Reply to root: Just insert it to the top (OK)
2. Reply to A': cannot just insert between A' and A'' (Ohhhh)
3. Reply to B or C: insert after B (OK)
4. Reply to A'': insert after A'' (OK)

So… Maybe we just consider case 1 and ignore the other cases. Or only ignore case 2.

The problem is the app does not have a banner to hinting the post result. Also, the [discussion](https://github.com/mastodon/mastodon-ios/issues/174) may related

PR #585, opened 2022-11-14
Fix authenticated user account not reloaded

kimar:
# Rationale

Currently the user profile won't be reloaded and will still reflect old user data (display name, avatar) e.g. when restarting / coming from background / using pull to refresh on the profile screen.

This PR fixes the behavior, now the user profile is reloaded when:

* the app is coming to foreground
* the user uses pull to refresh in their profile

This is easily extendible by emitting an event to `InstanceService.updateActiveUserAccountPublisher`.

# Demo

mastodon-profile-refresh-fix mov

### DIFF
--- a/MastodonSDK/Sources/MastodonCore/Service/InstanceService.swift
+++ b/MastodonSDK/Sources/MastodonCore/Service/InstanceService.swift
@@ -24,7 +24,8 @@ public final class InstanceService {
     weak var authenticationService: AuthenticationService?
     
     // output
-    
+    public let updateActiveUserAccountPublisher = PassthroughSubject<Void, Never>()
+
     init(
         apiService: APIService,
         authenticationService: AuthenticationService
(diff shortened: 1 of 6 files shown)


MastodonSDK/Sources/MastodonCore/Service/InstanceService.swift
```diff
@@ -24,7 +24,8 @@ public final class InstanceService {
     weak var authenticationService: AuthenticationService?
     
     // output
-    
+    public let updateActiveUserAccountPublisher = PassthroughSubject<Void, Never>()
```
kimar, line 27: I'm actually not entirely sure if the `InstanceService` is the best place to put this subject, but I thought it's most appropriate as the active user is tied to the instance.

MainasuK: Maybe `AuthenticationService`? 

The active user info is isolated to `AuthContext`. So that's fine to bind to any service that could access or handle `AuthContext`.

kimar: Yes, maybe that's better. 👍  I'll move it there and raise another PR as I've merged this one already.

Issue #504, opened 2022-11-04
[Feature] iOS Builtin Translation for Trööts


je1980:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Possibility to translate comments with iOS-built in translation e.g. „switch to German“ 

### Environment

```markdown
- Device:
- OS:
- Version:
- Build:
```

zeitschlag:
Hello @je1980,

thank you for that suggestion. Unfortunately, Apple doesn't provide an API to the Translate-App, so we can't solve this from the programming side of things. But what you could do instead, is this:

Select the text you want to have translated as if you want to copy it and then press the little arrows to the right until you see "Translate", like this:

https://user-images.githubusercontent.com/2580019/199965512-1148c192-13d1-44ae-a25c-20753161278d.MP4

("Übersetzen" is just the German translation for "Translate")

Does this solve the issue at hand? 🙂

david-coyle-sjc:
You can at least implement the Vision APIs for text recognition in images, that will offer translation when available.

j-f1:
I started looking into this but was turned away by the complexity of the code. There’s a translation endpoint `POST /api/v1/statuses/:id/translate` that will handle the translation if your server supports it, and `/api/v2/instance` includes a `configuration.translation.enabled` boolean value that tells you if support is available. I think it would make sense to add a new set of translation-related properties to the Status `NSManagedObject`, set them when the user requests a translation, and clear them when they hide it. (in particular, that would be the translated post content, the detected original language, and the translation provider).

Then a new UI would need to be wired up in `StatusView` to surface these controls, and handling for that button would need to be passed up the delegate chain.

zeitschlag:
Translation is implemented in #741 (if the server supports it)

PR #592, opened 2022-11-15
Add Live Text support to images

j-f1:
Live Text is now supported when viewing an image in full screen. Live Text only works on iOS 16 and newer. You can now tap on the image to hide the close button, status bar, and Live Text-related chrome. The long press gesture recognizer is suppressed on all of the Live Text chrome, and the tap gesture recognizer is suppressed on the links and buttons.

Related: #504 (confirmed that translation does indeed work with non-English text in images when my language is set to English)

https://user-images.githubusercontent.com/25517624/201817900-4ce3585c-6221-40f8-b9ce-d5eee6cfb6af.MP4


MainasuK:
The `Vision.framework` integration is awesome. Thanks a lot. 

I have thought about using saliency analysis to add auto-focus points for image attachments. And the new `Live Text` looks more useful.

MainasuK (approved):
Great work!

PR #598, opened 2022-11-15
Add punycode support to the server search box

### LINKED ISSUE #429: Internationalized Domain Names not supported

### Current Behavior

When searching for instances using IDNs the app cannot find it. 
Also user names with IDNs are not displayed correctly.

### Expected Behavior

Support IDN to allow searching for such domain names and display user names with IDNs correctly.

### Steps To Reproduce

When signing in search for example for the Mastodon instance „theländ.land“. This Instance exists, but is not found. 
Workaround: searching for „xn—thelnd-eua.land“.

### Environment

```markdown
All environments.
```

### Anything else?

_No response_

j-f1:
Fixes #429. I chose to only implement support for showing non-internationalized domains so I don’t have to wade into the issue of disambiguating forged domains (like `mаstodon.social` / `xn--mstodon-2fg.social`). Still open is the question of how IDN domains on the official list will be handled. Ideally they’d be displayed in a more human-friendly way (since a human has verified that they’re not a scam) but that will have to be handled later when it actually happens.


tigransimonyan:
why is this PR not merging?

PR #599, opened 2022-11-16
Limit single media to a reasonable height in feeds/threads

j-f1:
Ref: https://mastodon.online/@samhenrigold/109262994746866168

The previous behavior would allow images to be up to 10k pixels tall. Now, media items are compressed to 512px tall at most unless they’re the active item in the feed (i.e. `style == .plain`). I’m not sure if it makes sense to apply this new cap in all places, and if the number should be set lower. Thoughts? For context, you can still tap the image to open it full screen and zoom in that way.

I also changed the behavior of `MediaGridContainerView` to center a single media item when displaying it. (I’m not sure if multi-image posts can run into the same issue, since I think they’re constrained to a reasonable height already?)


j-f1:
Some more discussion on this from the Mastodon thread:

@samhenrigold:
> Thanks for doing this!! :ablobcatattention: From a visual standpoint, I recommend cropping (scale to fill) instead of letterboxing here. The thinking there is that it’ll be a trade off between legibility and seeing the full content, so we might as well show a little bit of legible content rather than seeing illegible-but-uncropped images.
> Further, that would let us do smart cropping using saliency but that’s wayyy down the road.

me:
> I tried cropping but (at least with the image in your post) the pic is only 512px high so it looked kinda bad at full height:
> 
> That said, I did find where the API returns the focal point of an image (manually inputted when uploading) so it would be theoretically possible to use that alongside scaling to fill.

PR #600, opened 2022-11-16
Remove redundant nil coalescing

woxtu:
Hi,

As the title says, this PR removes redundant nil coalescing to resolve compiler warnings.


kimar (approved):
Thank you @woxtu , LGTM 👍

PR #604, opened 2022-11-16
feat: Implement double-tap on profile Tab to quickly cycle through logged in Accounts

kimar:
# Rationale

Enable double-tap gesture on Profile-Tab to allow user to cycle through logged in accounts. Long-Press gesture is usable as well.

**I've pointed this PR towards `release-1.4.7`**

# Demo

https://user-images.githubusercontent.com/126418/202175062-bd6e7fe0-d67d-46fb-a927-f855607ed0af.mp4

### DIFF
--- a/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
+++ b/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
@@ -8,6 +8,7 @@
 import os.log
 import UIKit
 import Combine
+import CoreDataStack
 import SafariServices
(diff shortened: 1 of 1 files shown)


zeitschlag (approved):
Have you checked, that a11y works in regards to #520? Code-wise it looks good to me 👍

kimar:
@zeitschlag I just double-checked, you can still use the regular account switcher with voice over and it's long-press gesture, so the new double-tap gesture doesn't interfere with it.

PR #606, opened 2022-11-16
Fix timeline crash

MainasuK:
The lazy header setup code running on the network thread. The new SDK disallows that and also crashes the app. 

Dispatch it on the main thread to fix that issue.


kimar (approved):
🎉 LGTM

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -130,6 +130,7 @@ extension StatusView {
                                     authorization: authenticationBox.userAuthorization
                                 ).singleOutput()
                             }
+                            .receive(on: DispatchQueue.main)
                             .sink { completion in
                                 // do nothing
                             } receiveValue: { [weak self] response in
```
j-f1, line 136: I think this can be removed so it's just `.sink { [weak self] response in`

j-f1:
Also thanks for spotting that, I had noticed the crashes but couldn’t get to the bottom of them!

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
MainasuK: Thank you.

The Combine requires handling the `completion` case. Due to this event pipeline could emit an `Error`. So I cannot remove that.

j-f1: Ah, that makes sense!

PR #611, opened 2022-11-16
Fix parsing width/height for audio attachments

### LINKED ISSUE #414: Audio attachments are not shown

### Current Behavior

When viewing a post with an audio attachment, no audio player is shown. 

Additionally, there is no indication that the post has an audio attachment.

### Expected Behavior

When viewing a post with an audio attachment, an audio player is shown to allow the user to listen to the audio. 

Alternatively, there should at least be a message warning the user that they cannot listen to the audio attachment in the app, possibly with a link to open the post on the web so the web player can be used.

### Steps To Reproduce

1. In the iOS app, load a post in the home timeline (or post details screen) that has an audio attachment

### Environment

```markdown
- Device: iPhone 11 Pro
- OS: 15.4.1
- Version: 1.4.0
- Build: 121
```

### Anything else?

[Example post where I see the issue](https://mastodon.social/@rmrenner/108230206074707414)

My server is tech.lgbt, which is running Mastodon 3.5.1.

### LINKED ISSUE #642: [BUG] Audio media posted from web-app not available on iOS app

### Current Behavior

Posting an audio file from web-application won't show up on IOS app, but is playable from web-app. Text will be shown, media not.

See this post for example
https://layer8.space/@pd/109365014849444165

### Expected Behavior

_No response_

### Steps To Reproduce

1. Post an audio file from web-app, including image artwork.
2. Include Text
3. Open Mastadon iOS App and open your profile
4.

### Environment

```markdown
- Device: iPhone 11
- OS: 14.6
- Version: v1.4.5
- Build: 144
```

### Anything else?

_No response_

j-f1:
Fixes #414. Fixes #642. It seems like Mastodon no longer serves width/height on the `original` key in some/all cases. So I’m pulling the only available size (from `small`) instead.

### DIFF
--- a/MastodonSDK/Sources/MastodonCore/Extension/CoreDataStack/Status+Property.swift
+++ b/MastodonSDK/Sources/MastodonCore/Extension/CoreDataStack/Status+Property.swift
@@ -53,16 +53,29 @@ extension Mastodon.Entity.Status {
         let attachments = mediaAttachments.compactMap { media -> MastodonAttachment? in
             guard let kind = media.attachmentKind,
                   let meta = media.meta,
-                  let original = meta.original,
-                  let width = original.width,       // audio has width/height
-                  let height = original.height
+                  let original = meta.original
             else { return nil }
-            
+
+            let size: CGSize
+            if let width = original.width,
+               let height = original.height {
+                size = CGSize(width: width, height: height)
+                // audio may have the width/height on the “small” (thumbnail?) property instead
+            } else if
+                kind == .audio,
+                let small = meta.small,
+                let width = small.width,
+                let height = small.height {
+                size = CGSize(width: width, height: height)
+            } else {
+                return nil
+            }
+
             let durationMS: Int? = original.duration.flatMap { Int($0 * 1000) }
             return MastodonAttachment(
                 id: media.id,
                 kind: kind,
-                size: CGSize(width: width, height: height),
+                size: size,
                 focus: nil,    // TODO:
                 blurhash: media.blurhash,
                 assetURL: media.url,


tejuamirthi (approved):

MastodonSDK/Sources/MastodonCore/Extension/CoreDataStack/Status+Property.swift
```diff
@@ -53,16 +53,29 @@ extension Mastodon.Entity.Status {
...
+                // audio may have the width/height on the “small” (thumbnail?) property instead
+            } else if
+                kind == .audio,
+                let small = meta.small,
+                let width = small.width,
+                let height = small.height {
+                size = CGSize(width: width, height: height)
+            } else {
+                return nil
+            }
```
zeitschlag, line 72: This doesn't really help, if `meta.small` is nil.

```
(lldb) po media.remoteURL
▿ Optional<String>
  - some : "https://cdn.masto.host/mastodonart/media_attachments/files/109/302/787/076/266/355/original/04519903aafd999c.mp3"

(lldb) po meta
▿ Meta
  ▿ original : Optional<Format>
    ▿ some : Format
      - width : nil
      - height : nil
      - size : nil
      - aspect : nil
      - frameRate : nil
      ▿ duration : Optional<Double>
        - some : 205.688163
      ▿ bitrate : Optional<Int>
        - some : 162203
  - small : nil
  ▿ focus : Optional<Focus>
    ▿ some : Focus
      - x : 0.0
      - y : 0.0
  - length : nil
  - duration : nil
  - fps : nil
  - size : nil
  - width : nil
  - height : nil
  - aspect : nil
  - audioEncode : nil
  - audioBitrate : nil
  - audioChannels : nil
```

It's this [toot](https://mastodon.art/@simonstalenhag/109302790092203385) from the issue.

zeitschlag (changes requested):

j-f1:
Agreed, I think https://github.com/mastodon/mastodon-ios/pull/805 is the better solution here. Closing.

PR #612, opened 2022-11-17
Sort emoji alphabetically, and into sections

ReactiveRaven:
Sorts emoji alphabetically so they are easier for a user to scroll through, and similar content is likely to be closer together in the list (e.g. :ball_blue: :ball_red: :thumbs_down: :thumbs_up:).

Also groups emoji into sections as configured by the instance administrator.


kimar (approved):
Thank you very much @ReactiveRaven LGTM 👍

PR #616, opened 2022-11-17
Fix Video Player Constraints issue

painhapp:
When viewing videos some controls are overlapping the safe area insets.
For now it is sufficient to move the `didMove` to `viewDidLayoutSubviews`.
I would also like to enable PIP and auto rotation in an upcoming PR.

| before | after |
|--------|------|
|  Screenshot 2022-11-17 at 14 24 58 | Screenshot 2022-11-17 at 14 23 54 |

### DIFF
--- a/Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
+++ b/Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
@@ -45,7 +45,6 @@ extension MediaPreviewVideoViewController {
             playerViewController.view.widthAnchor.constraint(equalTo: view.widthAnchor),
             playerViewController.view.heightAnchor.constraint(equalTo: view.heightAnchor),
         ])
-        playerViewController.didMove(toParent: self)
         
         if let contentOverlayView = playerViewController.contentOverlayView {
             previewImageView.translatesAutoresizingMaskIntoConstraints = false
@@ -88,6 +87,12 @@ extension MediaPreviewVideoViewController {
                 }
                 .store(in: &disposeBag)
         }
+    }
+    
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        playerViewController.didMove(toParent: self)
     }
     
 }
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,7 +100,9 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
+  arm64-darwin-22
   x86_64-darwin-21
+  x86_64-darwin-22
 
 DEPENDENCIES
   arkana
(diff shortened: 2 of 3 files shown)


MainasuK:
Thank you for your contribution!

I actually have the plan to improve the video player. I consider just presenting an AVPlayerController in the modal. Just the same thing I did in the Twidere X [here](https://github.com/TwidereProject/TwidereX-iOS/blob/3fc202751edd3af251a526824c67ab745294928d/TwidereX/Protocol/Provider/DataSourceFacade%2BMedia.swift#L93-L114). And the rotation feature enables by default. Also, maybe easier PIP implementation could be.

painhapp:
> I actually have the plan to improve the video player. 

👍🏼 oh great, then I'll have a look at a different issue.

j-f1:
I don’t think Twitter supported mixing videos and photos until recently, but my concern with opening the player controller in full screen is that there wouldn’t be a way to swipe between media attachments.

MainasuK:
Make sense. 

We can fix the problem and see what to do that later.

Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
```diff
@@ -90,6 +89,12 @@ extension MediaPreviewVideoViewController {
         }
     }
     
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        playerViewController.didMove(toParent: self)
```
MainasuK, line 95: The layout changes on the desktop or iPad will raise this method multiple times. Maybe add a precondition check here?

MainasuK: Sorry. I try to understand this patch but no answer yet. Is this some UI magics?

@zeitschlag @kimar Please help me and give a review.

painhapp: Yes, you are right. 

From what I noticed during testing: It is indeed some kind of UI magic. The constraints itself are working, however the controls within `AVPlayerViewController` are not correctly set to the safe area insets.
Regarding the lifecycle: it also didn't work in `viewDidAppear`, the earliest trigger I found was `viewDidLayoutSubviews`. As you mentioned, it is run multiple times, but calling `didMove` multiple times wouldn't affect the behavior. It could be guarded with a flag, if you prefer that.
It would also be possible to keep it as it is and delay the `didMove`. Something like:
```
        DispatchQueue.main.asyncAfter(deadline: .now() + 1) { [weak self] in
            self?.playerViewController.didMove(toParent: self)
        }
```
Anyway, please let me know what you prefer and I'll adapt accordingly.

MainasuK: I guess the safe area not be applied until the VC attaches to a window. Whatever, it's should an SDK internal issue. If the multiple callings do not break anything. I think the current implementation is great. 

Thank you for your contributions!

PR #617, opened 2022-11-17
feat: Implement double-tap account switching on iPad

kimar:
# Rationale

Adds missing double-tap gesture for quick account switching to the iPad.

### DIFF
--- a/Mastodon/Extension/AppContext+NextAccount.swift
+++ b/Mastodon/Extension/AppContext+NextAccount.swift
@@ -1,0 +1,47 @@
+//
+//  AppContext+NextAccount.swift
+//  Mastodon
+//
+//  Created by Marcus Kida on 17.11.22.
+//
+
+import CoreData
+import CoreDataStack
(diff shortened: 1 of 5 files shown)


Mastodon/Extension/AppContext+NextAccount.swift
```diff
@@ -0,0 +1,47 @@
...
+        guard
+            let accounts = try? managedObjectContext.fetch(request),
+            accounts.count > 1
+        else { return nil }
+        
+        let nextSelectedAccountIndex: Int? = {
+            for (index, account) in accounts.enumerated() {
+                guard account == authContext.mastodonAuthenticationBox
+                    .authenticationRecord
+                    .object(in: managedObjectContext)
```
MainasuK, line 25: Here we could use the `objectID` for `account` to compare the same object without database access.

MainasuK (approved):

MainasuK:
LGTM

Issue #610, opened 2022-11-16
[DOCS] Add documentation for how to add new keys to model objects


j-f1:
I’ve started looking into several open issues (such as #396 #504 #591 #597) but they each require that some additional bits of information from the API need to be stored in the Core Data store (either as part of an existing model type or as an entirely new object). I’ve tried to do that but couldn’t figure out how. Could you add docs describing how to make modifications to the Core Data model?

PR #620, opened 2022-11-17
Add doc about how-it-works

MainasuK:
Try to explain something about the app.


MainasuK:
#610 

Hi, @j-f1. That's a draft wrote a long time ago. I try to finish it now with some extra info about Core Data. Maybe that is still not enough to clarify how the Core Data works. Just ask and I will answer whatever I could.

j-f1:
Thanks!

PR #622, opened 2022-11-17
[Refactor] Use UIView extension to create constraints (to parent view)

painhapp:
[Refactoring Task]
Creating anchors to a specific view (mostly the parent view) is reoccurring code, that can be achieved by an extension of UIView.
It easier to read - in case there is no special constraint necessary.


zeitschlag (approved):
LGTM

PR #579, opened 2022-11-13
Fix typos

woxtu:
Hi,

As the title says, this PR fixes some typos.

PR #623, opened 2022-11-17
Add spell checking using CSpell

j-f1:
This PR:
- adds documentation about how to enable spell checking in Xcode
- sets up an optional CSpell system for running spell checking locally that is also run as a CI step
  - I chose a Node.js-based tool here because I use it inside VS Code and it works great with camelCased names. Folks contributing don’t need to have Node.js installed unless they want to check spelling separately before pushing. (But they can also rely on the GitHub Action to give them feedback once they do push their code).
  - I looked for Swift-based alternatives but the only one I found, https://github.com/ezura/spell-checker-for-swift, seems to be unmaintained. I also wasn’t able to get it working properly when loaded in via Swift Package Manager (but I also did not really try very hard)
- Fixes all the detected typos: (closes #579)
  - changes to comments should be completely safe
  - changes to identifiers across the app should be fine too

However, there’s one thing I’m not sure of. I’ve renamed a few translation keys and I’d like to make sure that Crowdin is able to pick up those changes and not force translators to translate the affected keys again (since the translations should be identical). How can I do that? Or would it be best to leave the localization keys alone for now?

There’s also one key that is stored in Core Data and referenced a few other places that I’ve left alone for now but would like to sort out before merging this PR: `isActived`/`activedAt`. I’m not entirely sure what this means — @MainasuK could you explain what this means? Ideally we’d be able to rename this into something a bit clearer using a couple more words.


MainasuK:
The key is spell wrong and should be correct as `activeAt`. I want to use the same naming style likes `createdAt` or `updatedAt` but that should not be. I was aware of this mistake before but chose to leave it here. And if we follow the lightweight migration guide. Renaming a key is not a hard thing.

PR #625, opened 2022-11-17
Add names to the Run Script actions

j-f1:
This makes the Build Phases section a bit easier to navigate


zeitschlag (approved):
Awesome! I didn't know, that I can rename them. Thanks a lot!

PR #629, opened 2022-11-18
fix: Status notification shows "mentioned you" text

kimar:
# Rationale

Previously a status mention showed "mentioned you" in the Notifications area

# Demo

| Before | After |
|---|---|
| Bildschirm­foto 2022-11-18 um 10 52 59 | Bildschirm­foto 2022-11-18 um 10 53 57 |

### DIFF
--- a/Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
+++ b/Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
@@ -154,7 +154,7 @@ extension NotificationView {
                 )
             case .status:
                 self.viewModel.notificationIndicatorText = createMetaContent(
-                    text: L10n.Scene.Notification.NotificationDescription.mentionedYou,
+                    text: .empty,
                     emojis: emojis.asDictionary
                 )
             case ._other:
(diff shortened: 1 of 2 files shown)


Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
```diff
@@ -154,7 +154,7 @@ extension NotificationView {
                 )
             case .status:
                 self.viewModel.notificationIndicatorText = createMetaContent(
-                    text: L10n.Scene.Notification.NotificationDescription.mentionedYou,
+                    text: .empty,
```
zeitschlag, line 157: Why a string-extension and not just `String()` or `""`?
(Just a question, no need to change that :) )

zeitschlag (approved):

Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
kimar: I just have the feeling like this looks a bit more deliberate.

PR #631, opened 2022-11-18
Tag the build number automatically

MainasuK:
Tag the build number on the commit when deploying the TestFlight build.


.github/scripts/build-release.sh
```diff
@@ -36,6 +36,8 @@ BUILD_NUMBER=$(app-store-connect get-latest-testflight-build-number $ENV_APP_ID
 BUILD_NUMBER=$((BUILD_NUMBER+1))
 CURRENT_PROJECT_VERSION=${BUILD_NUMBER:-0}
 
+echo "GITHUB_TAG_NAME=build-$CURRENT_PROJECT_VERSION" >> $GITHUB_ENV
```
kimar, line 39: Just a minor thing, per default `echo XYZ` adds a newline at the end, if you want to suppress this you need to provide `-n` not sure if the newline is problematic, but I'd do:

 ```suggestion
echo -n "GITHUB_TAG_NAME=build-$CURRENT_PROJECT_VERSION" >> $GITHUB_ENV
```

kimar (approved):
LGTM just one minor comment you might or might not want to take. 👍

.github/scripts/build-release.sh
kimar: Actually in this context I think the newline won't matter as GitHub parses the env file.

PR #636, opened 2022-11-19
Update dependencies to latest versions

LumioseSil:
I figure it would be best to update our dependencies since we will be updating this app soon.


kimar (approved):
LGTM thank you @AtariDreams !

PR #662, opened 2022-11-21
Always allow to continue when reporting a post if at least one is selected

kimar:
# Rationale

Currently the "Report Post" functionality can be a bit misleading, as the next button is only enabled if an additional post is selected.

When selecting a user to block, not a post, then the flow will be the same and still show the Skip button (as no post has been selected previously).

Sorry for the person and their post depicted in the screenshot, this only serves as an example and has not been reported.

# Demo

| Before | After |
|---|---|
| RocketSim_Screenshot_iPhone_14_Pro_2022-11-21_11 28 30  | RocketSim_Recording_iPhone_14_Pro_2022-11-21_11 26 22|


j-f1:
I like this! I think it would also make sense to change the 2 buttons under “is there anything else we should know" into one big “Report” / “Submit” button to clarify that that is what’s gonna happen.

zeitschlag (approved):

kimar:
@j-f1 thank you, that's a good idea, Ive created #673 to track this and I'll merge this PR now.

Issue #663, opened 2022-11-21
[BUG] Search should not send each character to API while typing / Exhaustive database load


ralfr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When users type into the search field, every keystroke is send to the backend. Likely this is meant to enhance the user experience and provide a classical type-ahed search experience.

However, given the complexity of the database queries executed on the backed, this causes a ton of load on the backend side PostgreSQL database, which is essential for the entire performance of any instance.

We've seen the iOS app causing literally thousands of complex queries looking for "R", then for "RA" then for "RAL" then for "RALF" and so on.

### Expected Behavior

Type-ahead should be implemented in a way where, e.g.: The client waits until the user has at least provided three characters AND there was a typing pause of 800 ms or higher.

### Steps To Reproduce

n/a

### Environment

```markdown
n/a
```

j-f1:
At least on iOS, there’s currently throttling to once every 500ms. That said, it might make sense to increase it?

https://github.com/mastodon/mastodon-ios/blob/df5099dd30f1fc093143b4cc38edb7724d8d5c52/Mastodon/Scene/Search/SearchDetail/SearchDetailViewController.swift#L189

Yonodactyl:
I think that if we are introducing adjustments to the debounce, we should consider also adding some responsive feature like an `ActivityIndicator` or showing the `RefreshControl` on the Tableview. It would reinforce the idea behind #655

PR #667, opened 2022-11-21
Start search only with more than 3 characters

zeitschlag:
See #663


j-f1:
It would be nice if the user could hit the return key to explicitly perform a search shorter than 3 characters, if that’s what they actually wanted

Yonodactyl:
How would this handle, say pasting a link into the searchbar?

zeitschlag:
@Yonodactyl Good question, thank you. If the text is longer than those 3 characters, it would just do the search :)

zeitschlag:
@j-f1 That's a nice idea, ~~I'm struggling a bit on how to implement it reliably (I'm not asking for help here as I want to figure it out on my own).~~ It's implemented pretty hacky, but it works.

What I'm not sure about from a UX-view: Should tapping the search-key trigger search only once? On one hand, this feels weird and inconsistent to me, to be honest, as the search with more than three chars _just works_. On the other hand, with an automated search, this would still cause those database-queries with just one or two characters. What do you think?

I tend to allow search with more than three chars only.

idnovic:
I would like to still be able to search for my name "Ivo". 
Please make it >=3 or >2

kimar (approved):
LGTM 🚀

PR #682, opened 2022-11-22
Allow media viewer to be rotated to any orientation

### LINKED ISSUE #230: Ability to click on image and rotate to landscape mode (feature request)
It would be optimal to be able to select an image and then take advantage of landscape modeimage

### LINKED ISSUE #739: [Feature] Rotate images
When you tap on an image in a post and rotate the mobile, the image should be rotated too and should be displayed full screen.

j-f1:
Fixes #230. Fixes #739. I’m not 100% happy with it because the actual app is rotated to landscape as well (and then forced to rotate back to portrait once the overlay is dismissed) but it’s better than nothing!


Mastodon/Supporting Files/AppDelegate.swift
```diff
@@ -64,13 +64,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 }
 
 extension AppDelegate {
-    func application(_ application: UIApplication, supportedInterfaceOrientationsFor window: UIWindow?) -> UIInterfaceOrientationMask {
-        return UIDevice.current.userInterfaceIdiom == .phone ? .portrait : .all
-    }
-}
-
-extension AppDelegate {
-
```
zeitschlag, line 73: What happens due to this, is that several other screens get rotated as well, like the Onboarding or Safari:





Sure, one could just add this to every affected screen again and again: 

```swift
override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
    return UIDevice.current.userInterfaceIdiom == .phone ? .portrait : .all
}
```

I know: Rotation sucks. But I keep thinking: Is there a another way, a more elegant way to achieve rotation for one single screen and not change other existing ones? Like using a protocol? Or maybe a extra component, that determines (and returns orientation), if a screen should be rotated?

Something like [this](https://stackoverflow.com/a/44055120)?

zeitschlag (changes requested):
@j-f1 Thank you for tackling this.

As you say:

> the actual app is rotated to landscape [...] and then forced to rotate back to portrait

Do you have an idea, how to solve this? It's a bit annoying to me, but probably not an issue at all (and something I won't see that often, as I use my phone in portrait only).

j-f1:
One way could be to handle rotation manually by reading from the gyros/accelerometer (ugh). Another could be to overlay a snapshot of the previous view controller until the rotation finishes (not sure if that would be effective though)

Mastodon/Supporting Files/AppDelegate.swift
j-f1: Maybe! (Or it could go the other way and have a protocol that implements that method to constrain orientations for us). That said I think at least the Safari change is an improvement. Sometimes websites (especially ones with videos) need to be viewed in landscape.

zeitschlag: Regarding Safari: Fair point, yes, I'm convinced 👍  It just changed something, I wasn't  sure, if this was intended, as the app itself is in portrait only (on phone). But the broken onboarding is still an issue :)

Reading from the sensor or using an overlay to fake something feels either pretty overkill or pretty hacky (or both). If there's no better way, then that's what it is :D

Gargron:
I think we probably don't want the whole app to rotate, and we don't really need landscape for pictures. I think we should support landscape orientation for full-screen video, and that's it. Is that more doable?

j-f1:
Doing it for both images and video is, I think, easier than just doing it for video. I’m planning to finish walking through the app’s screens tomorrow to find anywhere else where it could be advantageous to go landscape (unlikely) or these changes erroneously let you go landscape (found one so far, there might be or two more).

j-f1:
I did a check through the app and I think those changes should be all that’s needed ^

Mastodon/Scene/Compose/ComposeViewController.swift
```diff
@@ -129,7 +129,7 @@ extension ComposeViewController {
 extension ComposeViewController {
   
     private func showDismissConfirmAlertController() {
-        let alertController = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
+        let alertController = PortraitAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
```
zeitschlag, line 132: What's the reason you use this `PortraitAlertController` only here?

Rest looks good to me, if we want to support rotation for media. Is it possible to have it for video only, too? To be honest: I didn't look into this, but there's at least `MastodonAttachment.Kind`. Maybe that's helpful?

j-f1: ¯\\\_(ツ)\_/¯ I have no idea why it’s necessary, but maybe it has something to do with the action sheet presentation? That isn’t used anywhere else in the app, and the alerts presented as an alert don’t cause any rotation problems.

> Rest looks good to me, if we want to support rotation for media. Is it possible to have it for video only, too? To be honest: I didn't look into this, but there's at least `MastodonAttachment.Kind`. Maybe that's helpful?

Theoretically yes, but sometimes people post landscape images so I think it’s good to always offer the option. Also there would be some weirdness when swiping between image and video attachments in the same post if you were in landscape.

zeitschlag: Good point, thank you!

zeitschlag (approved):

PR #684, opened 2022-11-23
Fix share extension reentry crash issue

### LINKED ISSUE #676: [BUG] Share extension often fails to load first time

### Current Behavior

About half the time the share extension fails to load, as shown in the video – it has, so far, always loaded on the second attempt.

https://user-images.githubusercontent.com/530912/203305146-0fc4ddac-7559-428e-ae95-63685e82f942.MOV

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- Device:
- OS: iOS
- Version: 16.2b
- Build:
```

### Anything else?

_No response_

MainasuK:
The CoreDataStack instance somehow leaked when the share extension exited. So if extension reentry before killed by the system. The duplicated init CoreDataStack raises an exception and crashes the extension. 

Using a singleton AppContext to avoid the CoreDataStack leaking to workaround this problem.

#### Before
https://user-images.githubusercontent.com/7940186/203484270-c582e4cd-b573-41f2-b3cd-d4dfe1b1c795.mp4

#### After
https://user-images.githubusercontent.com/7940186/203484274-13652c9a-8eae-4963-93c3-6f5c0e8737e9.mp4

### DIFF
--- a/MastodonSDK/Sources/MastodonCore/AppContext.swift
+++ b/MastodonSDK/Sources/MastodonCore/AppContext.swift
@@ -115,6 +115,10 @@ public class AppContext: ObservableObject {
             .store(in: &disposeBag)
     }
     
+    deinit {
+        os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s", ((#file as NSString).lastPathComponent), #line, #function)
+    }
+    
 }
 
 extension AppContext {
--- a/ShareActionExtension/Scene/ShareViewController.swift
+++ b/ShareActionExtension/Scene/ShareViewController.swift
@@ -21,7 +21,7 @@ final class ShareViewController: UIViewController {
     
     var disposeBag = Set<AnyCancellable>()
     
-    let context = AppContext()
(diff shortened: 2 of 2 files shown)


kimar (approved):
I'm wondering if we should always use `AppContext.shared` and make the initializer private? `AppDelegate` is also instantiating `AppContext`.

MainasuK:
@kimar Making it as a singleton class is make sense. But I'm not sure to create a singleton in the package is a good idea. It's will force the app extension (a.k.a `NotificationService`) to init it but not actually in use.

j-f1:
I'm pretty sure static variables are always lazy so if you never access them they will never be initialized

MainasuK:
> I'm pretty sure static variables are always lazy so if you never access them they will never be initialized

Yes. It's should be a lazy init object. Another reason is I try to not use singleton too much. And it's already is a shared object in the `Mastodon` app target. Not a class static variable but a getter on shared `AppDelegate`.

PR #686, opened 2022-11-24
Delete CoreData records when user is blocked / muted from outside the App

### LINKED ISSUE #626: Blocked/muted users still show up in global feed
Example. A user I have both blocked and muted still shows.

[6B5D22F9-4AA3-4833-84BD-B0F4C1A80AC8](https://user-images.githubusercontent.com/11892959/202633741-57999861-422d-4351-b5e3-207b489fddb6.jpeg)

kimar:
# Rationale

As a user I want blocked users, which I have blocked from outside the App, to disappear from my timelines within the iOS App.

Resolves #626 

# How to test

* Log in to the iOS App
* Log in to Web (or another App)
* Mute / Block a user
  * Open the iOS App from Background / pull to refresh the timeline
  * Verify posts from the muted / blocked user have disappeared

@MainasuK / @zeitschlag could you please give this some QA? Thx!


MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
```diff
@@ -21,6 +21,36 @@ extension APIService {
...
+        let managedObjectContext = backgroundManagedObjectContext
+        
+        let response = try await Mastodon.API.Account.mutes(
+            session: session,
+            domain: authenticationBox.domain,
+            authorization: authenticationBox.userAuthorization
+        ).singleOutput()
+        
+        let userIDs = response.value.map { $0.id }
+        let predicate = NSPredicate(format: "%K IN %@", #keyPath(MastodonUser.id), userIDs)
```
MainasuK: Please use the `MastodonUser.predicate(domain:ids:)` as predicate here. That will guarantee all users locate in specific domain. Users in different domain may use the same ID.

MastodonSDK/Sources/CoreDataStack/CoreData.xcdatamodeld/CoreData 5.xcdatamodel/contents
```diff
@@ -0,0 +1,254 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
MainasuK: It's looks like no difference between version 4 and 5. 🤔

MainasuK: Oh, I see. The Delete Rules changes.

MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+Account+Friendship.swift
```diff
@@ -215,6 +215,53 @@ extension Mastodon.API.Account {
...
+            .tryMap { data, response in
+                let value = try Mastodon.API.decode(type: [Mastodon.Entity.Account].self, from: data, response: response)
+                return Mastodon.Response.Content(value: value, response: response)
+            }
+            .eraseToAnyPublisher()
+    }
+    
+    private struct BlocksQuery: GetQuery {
+        var queryItems: [URLQueryItem]? {
+            [URLQueryItem(name: "limit", value: "-1")]
```
MainasuK: I haven't tried this API. And could we grab all block users when setting the limit to "-1"?

MainasuK (commented):
We could merge it first and make a test then decide to release it or later. 

Accessing a deleted CoreData object will crash the app. Besides using objects from `NSFetchedResultsController`. The app needs to observe the object delete status or just use the object via `ObjectID` and make sure it is valid.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
```diff
@@ -21,6 +21,36 @@ extension APIService {
...
+        
+        let fetchRequest = MastodonUser.fetchRequest()
+        fetchRequest.predicate = predicate
+        fetchRequest.includesPropertyValues = false
+        
+        try await managedObjectContext.performChanges {
+            let accounts = try managedObjectContext.fetch(fetchRequest) as! [MastodonUser]
+            
+            for account in accounts {
+                managedObjectContext.delete(account)
```
j-f1: How would this interact with manually viewing a blocked/muted user’s profile?

MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+Account+Friendship.swift
DK1DA: Oh no, this was just for testing purposes and I'll remove it, the backend actually converts it to an absolute number, so we have a limit of 1 in this case. To have a proper implementation we probably need to implement pagination here.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
DK1DA: I'Ve update this.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
DK1DA: | Text | Screenshot |
|---|---|
| @j-f1 when viewing a blocked user's profile it would appear like this in the app. That's already been implemented before though.|RocketSim_Screenshot_iPhone_14_Pro_2022-11-24_13 28 53|

j-f1: Ok, just wanted to make sure deleting the user record from the DB doesn’t have any negative consequences for this experience. (But also: would it make sense to only delete the posts from Core Data and not the actual user record?)

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Block.swift
```diff
@@ -22,6 +22,36 @@ extension APIService {
...
+        let managedObjectContext = backgroundManagedObjectContext
+        
+        let response = try await Mastodon.API.Account.blocks(
+            session: session,
+            domain: authenticationBox.domain,
+            authorization: authenticationBox.userAuthorization
+        ).singleOutput()
+        
+        let userIDs = response.value.map { $0.id }
+        let predicate = NSPredicate(format: "%K IN %@", #keyPath(MastodonUser.id), userIDs)
```
MainasuK: Maybe missing This.
Same to https://github.com/mastodon/mastodon-ios/pull/686#discussion_r1031330565

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
DK1DA: @j-f1 hm that's a good question, what would you say @MainasuK ?

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Block.swift
DK1DA: I also noticed it, I fixed this with my latest commit. Thx 👍

kimar:
@MainasuK I've introduced `Link: ` header pagination for block / mute list and recursion the function iterates through those lists with my latest commit: https://github.com/mastodon/mastodon-ios/pull/686/commits/c6ac9481b0fce08bafeb85b78681249363eb7efd

Obviously this is tricky to test and I'm not sure what happens if there is a user with thousands of people on their blocklist, this might likely become a problem then.

I'm also thinking if it might not be sufficient to fetch the latest X (e.g. 100) users on the blocklist, then erase their local data. This would mean that, if the user just open the app once while blocking 100 users they should be fine.

What do you think?

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
MainasuK: 🤔 Interesting… I try to block random people and enter Home Screen. Then active the app. The app just display that profile in normal way. Whatever, the good news is the app not crash.  

Because the data clean happened when app active with network requests. So it's maybe delay to some point user is skimming the data will be removed. Or the data on screen when user go back from home is marked deleted.

j-f1:
Maybe you could request all blocks since the most recent saved block? (not sure how unblocking would be handled though)

MainasuK:
@kimar

If the block list return in marked time order. It’s make sense only clean recent N blocked users from API. If you worry about batch of new blocked users. We could store the blocked userID list. Then we not fetch next page until no new ID coming. 

And I think the timeline will not gives blocked users’ posts. So maybe we could clean DB when user do block action AND when app go background (or foreground).

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
MainasuK: I agree with @j-f1. We only needs clean the posts from home timeline (only the index, not the post itself) and notification timeline (only the index, same above). (Any others? I think we only persist these two timeline.)

So we need find user then use reverse relationship to find the post then feed index. And please be care of the `reblogged` relationship. :)

kimar:
@j-f1 unfortunately I can only query blocks / mutes and then go back and forward from there, but I can't easily start at any given point without first doing an initial request and fetching the next / previous pointer.

@MainasuK I'Ve just checked and the results of the block list were ordered chronologically by order of blocking the accounts. So I'm removing the recursion for now and also I've decided to use the server limit.

Regarding the timeline: For me it was still showing blocked user's posts / boost before I deleted them. But I'll do a cleanup when a user performs the block additionally.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Mute.swift
MainasuK: Another way to find the feed index is use compound predicate with relationship condition. ~That will be faster.~

DK1DA: @MainasuK I've just pushed a commit which does the data purge on mute/block action. I'm currently not sure what you mean cleaning the posts from the home timeline by index. Maybe you could, if you have some time, give me pointer or make an adjustment in this branch? This sounds like we could possibly also get rid of the CoreData model changes?

MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+Account+Friendship.swift
```diff
@@ -215,6 +215,70 @@ extension Mastodon.API.Account {
...
+    /// - Version: 3.3.0
+    /// # Last Update
+    ///   2021/4/1
+    /// # Reference
+    ///   [Document](https://docs.joinmastodon.org/methods/blocks/)
+    /// - Parameters:
+    ///   - session: `URLSession`
+    ///   - domain: Mastodon instance domain. e.g. "example.com"
+    ///   - authorization: User token.
+    /// - Returns: `AnyPublisher` contains `Relationship` nested in the response
```
zeitschlag, line 238: Awesome! Documentation! 🥳

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Block.swift
```diff
@@ -110,3 +149,11 @@ extension APIService {
     }
     
 }
+
+extension Set<Status> {
+    func deleteAllFeedsForBlockOrMute(in managedObjectContext: NSManagedObjectContext) {
+        map { $0.feeds.union($0.reblogFrom.map { $0.feeds }.flatMap { $0 }) }
+        .flatMap { $0 }
+        .forEach(managedObjectContext.delete)
+    }
```
MainasuK: Cool. The home timeline feeds now deleted immediately. 

And I'm sorry my explain not cover all cases. Another timeline kind is notification. 
The Notification has two relationship: 
- `.account`
- `.status`: a related status if notification gives

So there are two paths:
1.  `MastodonUser.statuses` ->  (`Status.reblogFrom` ->) `Status.notifications`-> `Notification.feeds`
2. `MastodonUser.notifications`

DK1DA: Ok got it, I've updated the PR. I have to take a break from this now as my brain is slowly melting :-D Maybe we can try out and test the current implementation as-is and improve with a compound predicate later?

MainasuK (commented):
LGTM !

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Block.swift
MainasuK: It's depends on the size of database. I guess use reverse relationship will be always the best choice.

MainasuK (approved):

PR #697, opened 2022-11-26
Better onboarding

### LINKED ISSUE #690: Improved Onboarding
- Next-Button as NavigationItem, "normal" Back-button
- Education
- Pick Server (automatically based on language)
- Privacy Server/App
- Password-Confirmation (~~+Password AutoFill~~ we would need associated domains, which, when it comes to a decentralized social network, is a bit tricky.)
- Resend-Link
- Consider closed servers for login-autocomplete, too

zeitschlag:
see #690


Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentPageView.swift
```diff
@@ -0,0 +1,59 @@
+//
+//  WelcomeContentPageView.swift
+//  Mastodon
+//
+//  Created by Nathan Mattes on 26.11.22.
```
nemobis: Please use  the yyyy-mm-dd format.
https://xkcd.com/1179/

Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentPage.swift
```diff
@@ -0,0 +1,49 @@
...
+//  WelcomeContentPage.swift
+//  Mastodon
+//
+//  Created by Nathan Mattes on 26.11.22.
+//
+
+import UIKit
+import MastodonLocalization
+
+enum WelcomeContentPage: CaseIterable {
```
DK1DA, line 12: As this enum is, at least currently, only used to switch over different content with the same layout, I'm wondering if we should make this more generic e.g. `pageOne`, `pageTwo`, etc. then we'd only need to update the i18n values later on, if we'd like to change the onboarding content.

Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentPageView.swift
DK1DA: We could also consider dropping the header in general. In its current form I don't see any value in it at all. The filename is part of, well, the filename. And the date of creation can be inferred from the Git history.

nemobis: Definitely. I'm not sure what the status of copyright assignment is in this project, but *if* you want to have file-level headers this is the recommended way: https://reuse.software/faq/#step-2

Localization/StringsConvertor/input/Base.lproj/app.json
```diff
@@ -287,9 +300,19 @@
                 "no_results": "No results"
             }
         },
+        "privacy": {
+            "title": "Privacy",
+            "policy": {
+                "ios": "Privacy Policy - Mastodon for iOS";
+                "server" = "Privacy Policy - %s";
```
DK1DA, line 307: There is still some invalid JSON here.

kimar (approved):
LGTM besides this one broken JSON 👏  Great work!

Localization/StringsConvertor/input/Base.lproj/app.json
zeitschlag: Ah, shit, thank you for fixing it!

PR #701, opened 2022-11-27
Fix scroll to top when tab is tapped

kylebshr:
The current logic always scrolls to top, even if the view controller isn't on top and the navigation controller is popping, because by the time `didSelect` is called the navigation controller state is already updated.


kimar (approved):
LGTM 👍

PR #702, opened 2022-11-27
Space buttons instead of stretching

kylebshr:
The tap targets for buttons are currently misleading, and I'll tap the space in between to go to the details and end up performing an unintended action. This is how Twitter works, so users migrating will probably expect this too.


kimar (approved):
👍  Thank you!

PR #704, opened 2022-11-28
Add support for server-side user preferences

### LINKED ISSUE #297: Show all media setting
Please allow us to show all sensitive media by default/the ability to disable the blur on all posts.

### LINKED ISSUE #331: The preferences is ignored by the application
App make the correct behaviors from preferences settings.

Document:
https://docs.joinmastodon.org/methods/accounts/preferences/

### LINKED ISSUE #472: When profile is set to require follow requests the default visibility is always followers only
This should not be tied to follow requests being enabled, this should be its own thing to set default visibility

### LINKED ISSUE #511: [BUG] <title> Always expand posts marked with content warnings not working

### Current Behavior

In Settings I set both Media Display: Always show media and Always expand posts marked with content warnings to ON, then I SAVE CHANGES, then cancel out of settings.

Content with warnings still ask me to Tap to reveal.

### Expected Behavior

Content with warnings should be displayed.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Device: iPhone and iPad 
- OS: 16
- Version:
- Build:
```

### Anything else?

The correct behaviour is taken in the Toot! app

### LINKED ISSUE #750: Add setting for default message scope
... or alternatively honor the user's setting from their account settings on the server.

For some reason, the app always selects "Followers only" as the default posting scope for me. It might be because I have follow requests enabled, but for me, those are two completely separate things. I want my posts to be public by default, but using the Mastodon app, I always need to be careful to change it before sending the post. This is especially annoying as Mastodon (not this app in particular) doesn't let you change the scope after the post is published, so the only option is to delete and repost.

Related to #236 as their conclusion "make this configurable" is the same.

j-f1:
Fixes #331.
Preferences currently returned by Mastodon (v4.1.0):

- `posting:default:visibility`: configures the default visibility value for new posts (except that replies to private/DM posts will inherit that visibility). Maybe fixes #472, fixes #750.
- `posting:default:sensitive`: _not supported_. The compose UI does not allow configuring sensitive status for media separately from adding a content warning.
- `posting:default:language`: controls the default language used when composing a new post. Also, this language will always be available as the first element of the language menu.
- `reading:expand:media`: controls when media is hidden and controls visibility of the button icon as needed (e.g. it will always show up on posts with media if you set media to always be hidden by default). Fixes #297.
- `reading:expand:spoilers`: if `true`, content warnings will never be shown. (note: if `reading:expand:media` is not set to `show_all`, people won’t be able to see the content warning before viewing media. Should this be changed?). Fixes #511.
- `reading:autoplay:gifs`: _not supported_. Will implement after #815 is merged.

Preferences are re-fetched every time you switch accounts and every time the app is brought into the foreground. Unsure if this request frequency should be reduced. Thoughts? The API endpoint doesn’t seem terribly heavy on the backend, but I could imagine a bunch of folks switching back and forth from the app quickly causing a bunch of server overload. It is nice that changes to preferences are quickly picked up though!

### DIFF
--- a/Documentation/How-it-works.md
+++ b/Documentation/How-it-works.md
@@ -40,9 +40,9 @@ There is a self-hosted Swift Package that contains the common libraries to build
 #### CoreDataStack
 App uses Core Data as the backend to persist all entitles from the server. So the app has the capability to keep the timeline and notifications. Another reason for using a database is it makes the app could respond to entity changes between different sources. For example, a user could skim in the home timeline and then interact with the same post on other pages with favorite or reblog action. Core Data will handle the property modifications and notify the home timeline to update the view.
 
-To simplify the database operations. There is only one persistent store for all accounts. We use `domain` to identify entity for different servers (a.k.a instance). Do not mix the `domain` with the Mastodon remote server name. For example. The domain is `mastodon.online` whereever the post (e.g. post come from `mstdn.jp`) and friends from for the account sign in `mastodon.online`. Also, do not only rely on `id` because it has conflict potential between different `domain`. The unique predicate is `domain` + `id`.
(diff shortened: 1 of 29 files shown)


j-f1:
Updated to add support for `posting:default:language`!

PR #706, opened 2022-11-29
fix: Text in compose post is not selectable, focussable, pastable

### LINKED ISSUE #695: [BUG] Share Sheet: Text cannot not scrolled in, selected, copied or pasted

### Current Behavior

When using the Share Sheet to create a post, the text inside the share sheet cannot be scrolled, selected, copied or pasted.

### Expected Behavior

Selecting text, navigating text, copying and pasting is possible.

### Steps To Reproduce

1. Select a photo in Photos.ap
2. Share -> Mastodon
3. Enter text into the compose post share sheet
4. Try to navigate text, try to copy, try to paste something

### Environment

```markdown
- Device: iPhone 14 Pro
- OS: 16.1.1
- Version: 1.4.7
- Build: 199
```

### Anything else?

_No response_

kimar:
# Rationale

When adding a **portrait format** image attachment, the text input in the compose view cannot be interacted with anymore. This is due to the image overlaying the input (invisibly). Due to the `clipShape` modifier we're applying, this is not visible but still the image view is capturing touches, making it impossible to interact with the text input, this is true for bot the in-app compose as well as the share extension as they both share the same underlying components.

When removing the clipping you can see the image cover the text input.

IMG_1512

This PR dibbles hit testing on the image, so it will forward touches to its parent(s). As there is no direct interaction with the image anyways I think this is a reasonable solution.

Fixes #695

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Attachment/AttachmentView.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Attachment/AttachmentView.swift
@@ -30,6 +30,7 @@ public struct AttachmentView: View {
                     Image(uiImage: image)
                         .resizable()
                         .aspectRatio(contentMode: .fill)
+                        .allowsHitTesting(false)
                 }
             )
             .overlay(


MainasuK (approved):
Good catch!

PR #709, opened 2022-11-30
Enable multiline poll options

### LINKED ISSUE #291: The text of the polls's choices is cut off
Just noticed that the text of the poll's choices is cut off, it is not shown in full. 
Compare both images, one from the app and the other from de browser.
Mastodon_app
Browser

nossipova:
Fixes #291.

The text field has been replaced with a text view. There are still some glitches in view resizing, because everything has a dynamic height now, e.g. the first meta text sometimes appears only after scrolling. I'm still trying to find out how to improve it, but I'm not sure that it can be achieved without rewriting the relevant parts in SwiftUI.


nossipova:
It seems to work now. For a better scrolling performance the status view and its subviews need to be rewritten in SwiftUI to avoid multiple auto layout passes, but that would be to much for this PR.

PR #717, opened 2022-12-01
Workaround paste crash on iOS 14.x issue

### LINKED ISSUE #660: Crashes when I paste
Whenever I copy text to the clipboard and then start a new post and hold to reveal the context menu and select the Paste option, the application crashes immediately.

MainasuK:
Resolve #660 

Paste anything in post compose scene on iOS 14.5+ will crash the app. However, it's normal on iOS 15.0+ and iOS 16.0+. It looks like the paste image feature intro in #486 is somehow not working on iOS 14.x. 

Just disable it for iOS 14.x to work around the crash problem.

#### Reproduce the issue on iPad mini 5 with iOS 14.8.

image

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Vendor/MetaTextView+PasteExtensions.swift
+++ b/MastodonSDK/Sources/MastodonUI/Vendor/MetaTextView+PasteExtensions.swift
@@ -13,17 +13,21 @@ extension MetaTextView {
     public override func paste(_ sender: Any?) {
         super.paste(sender)
         
-        var nextResponder = self.next;
-        
-        // Force the event to bubble through ALL responders
-        // This is a workaround as somewhere down the chain the paste event gets eaten
-        while (nextResponder != nil) {
-            if let nextResponder = nextResponder {
-                if (nextResponder.responds(to: #selector(UIResponderStandardEditActions.paste(_:)))) {
-                    nextResponder.perform(#selector(UIResponderStandardEditActions.paste(_:)), with: sender)
+        // fix #660
+        // https://github.com/mastodon/mastodon-ios/issues/660
+        if #available(iOS 15.0, *) {
+            var nextResponder = self.next;
+            
+            // Force the event to bubble through ALL responders
+            // This is a workaround as somewhere down the chain the paste event gets eaten
+            while (nextResponder != nil) {
+                if let nextResponder = nextResponder {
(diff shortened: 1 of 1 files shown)


MastodonSDK/Sources/MastodonUI/Vendor/MetaTextView+PasteExtensions.swift
```diff
@@ -13,17 +13,21 @@ extension MetaTextView {
...
-        
-        // Force the event to bubble through ALL responders
-        // This is a workaround as somewhere down the chain the paste event gets eaten
-        while (nextResponder != nil) {
-            if let nextResponder = nextResponder {
-                if (nextResponder.responds(to: #selector(UIResponderStandardEditActions.paste(_:)))) {
-                    nextResponder.perform(#selector(UIResponderStandardEditActions.paste(_:)), with: sender)
+        // fix #660
+        // https://github.com/mastodon/mastodon-ios/issues/660
+        if #available(iOS 15.0, *) {
```
zeitschlag, line 18: Instead of enabling this one feature for iOS >15, we could also think about supporting iOS >15 for the entire app, so just the two most recent iOS-versions. What do you think? cc @kimar @Gargron

j-f1: That could be cool! iOS 15 also lets us use things like `AsyncPublisher` to do interop between async sequences and Combine.

DK1DA: 👍 on iOS >= 15 for me as well. I like to pursue n-1 for supported OSes, and given iOS 15 is supported by such a large percentage of device, I personally don't s a reason not to bump.

j-f1: (iOS 15 supports all devices that iOS 14 supports IIRC)

Gargron: Sounds good to me

MainasuK: If we update the minimal version in the next build. The iOS 14.x still could install the app but only for v1.4.7. 

Maybe we fix this bug and push as v1.4.8. Then update the min OS in v1.5.0. So this issue not bother old OS users.

PR #718, opened 2022-12-01
Fix timeline reload blink issue

### LINKED ISSUE #714: [BUG] Timeline jumps erratically when refreshing

### Current Behavior

When the home timeline content refreshes, there is a visual stutter/jump which is disturbing. The application does not lose the current position on the timeline, but for a brief moment it jumps back and forth in an unsettling manner.

### Expected Behavior

The timeline should only ever move in response to user input, never from background refresh or other activity.

### Steps To Reproduce

1. View the home tab timeline
2. Trigger a refresh of the timeline which does not effect the position of the timeline (such as posting a new status)
3. Witness the timeline jumping erratically in an unsettling manner. you can see a flash between 7 and 8 seconds on the attached video. (attached as zip for size limits)

[Screen Recording 2022-11-30 at 1.42.18 PM.zip](https://github.com/mastodon/mastodon-ios/files/10126307/Screen.Recording.2022-11-30.at.1.42.18.PM.zip)

### Environment

```markdown
- Device: M1 MacBook Air
- OS: macOS 13.0.1
- Version: Mastodon for iOS v1.4.7
- Build: (199)
```

### Anything else?

This jarring behavior can be seen any time the timeline content refreshes without human triggering.

MainasuK:
resolve #714 

The new `async/await` style `Task` grouped task default is not on the main queue. So the UI will blink due to tableView reload job trigger context switch.

 Force the `Task` on the `MainActor` to reduce the context switch to fix the blink problem.

### DIFF
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+Diffable.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+Diffable.swift
@@ -41,7 +41,7 @@ extension HomeTimelineViewModel {
                 guard let self = self else { return }
                 guard let diffableDataSource = self.diffableDataSource else { return }
                 self.logger.log(level: .debug, "\((#file as NSString).lastPathComponent, privacy: .public)[\(#line, privacy: .public)], \(#function, privacy: .public): incoming \(records.count) objects")
-                Task {
+                Task { @MainActor in
                     let start = CACurrentMediaTime()
                     defer {
                         let end = CACurrentMediaTime()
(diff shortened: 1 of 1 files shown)


j-f1:
How expensive is this task? Would it make sense to at least run the starting part where the data source update is constructed in the background?

kimar (approved):
LGTM 👍

MainasuK:
> How expensive is this task? Would it make sense to at least run the starting part where the data source update is constructed in the background?

```
2022-12-02 00:24:59.551544+0800 Mastodon[2047:474099] [ViewModel] HomeTimelineViewModel+Diffable.swift[43], setupDiffableDataSource(tableView:statusTableViewCellDelegate:timelineMiddleLoaderTableViewCellDelegate:): incoming 5941 objects
2022-12-02 00:24:59.665264+0800 Mastodon[2047:474099] [ViewModel] HomeTimelineViewModel+Diffable.swift[98], setupDiffableDataSource(tableView:statusTableViewCellDelegate:timelineMiddleLoaderTableViewCellDelegate:): snapshot has changes
2022-12-02 00:24:59.794127+0800 Mastodon[2047:474099] [ViewModel] HomeTimelineViewModel+Diffable.swift[118], setupDiffableDataSource(tableView:statusTableViewCellDelegate:timelineMiddleLoaderTableViewCellDelegate:): applied new snapshot
2022-12-02 00:24:59.794292+0800 Mastodon[2047:474099] [ViewModel] HomeTimelineViewModel+Diffable.swift[48], setupDiffableDataSource(tableView:statusTableViewCellDelegate:timelineMiddleLoaderTableViewCellDelegate:): cost 0.2423s to process 5941 feeds
```

The CoreData fetch job is already dispatched in the background. The UI calculation and modification need to be worked on the main thread. I think the `newSnapshot` init won't cost too much. So It's should be safe.

Maybe there is another way to hold the `contentOffset` and reload the data source. Also, I hope the iOS SDK could give a choice to keep position when applying snapshot like what `AsyncDisplayKit` did.

j-f1:
Ok, that makes sense! Thanks for getting that number :)

PR #736, opened 2022-12-06
Add Debug Option For Displaying Touches

cdoncarroll:
When reporting a bug or pitching a new feature, it can be very helpful to see just how the user is interacting with the UI. This PR adds the functionality to do just that by providing a new option in the debug menu to toggle on/off the visibility of touches.

### Example
https://user-images.githubusercontent.com/5335670/206009337-01fc352f-2266-49cd-9348-ff8ee3250694.mp4

### DIFF
--- a/Mastodon/Extension/UIApplication.swift
+++ b/Mastodon/Extension/UIApplication.swift
@@ -22,5 +22,13 @@ extension UIApplication {
 
         return version == build ? "v\(version)" : "v\(version) (\(build))"
     }
+    
+    func getKeyWindow() -> UIWindow? {
+        return UIApplication
+            .shared
+            .connectedScenes
+            .flatMap { ($0 as? UIWindowScene)?.windows ?? [] }
+            .first { $0.isKeyWindow }
+    }
 
 }
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DebugAction.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DebugAction.swift
@@ -122,6 +122,10 @@ extension HomeTimelineViewController {
             identifier: nil,
             options: [],
             children: [
+                UIAction(title: "Toggle Visible Touches", image: UIImage(systemName: "hand.tap"), attributes: []) { [weak self] action in
+                    guard let window = UIApplication.shared.keyWindow as? TouchesVisibleWindow else { return }
+                    window.touchesVisible = !window.touchesVisible
+                },
                 UIAction(title: "Toggle EmptyView", image: UIImage(systemName: "clear"), attributes: []) { [weak self] action in
                     guard let self = self else { return }
                     if self.emptyView.superview != nil {
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DebugAction.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DebugAction.swift
@@ -122,8 +122,8 @@ extension HomeTimelineViewController {
             identifier: nil,
             options: [],
             children: [
-                UIAction(title: "Toggle Visible Touches", image: UIImage(systemName: "hand.tap"), attributes: []) { [weak self] action in
-                    guard let window = UIApplication.shared.keyWindow as? TouchesVisibleWindow else { return }
+                UIAction(title: "Toggle Visible Touches", image: UIImage(systemName: "hand.tap"), attributes: []) { _ in
+                    guard let window = UIApplication.shared.getKeyWindow() as? TouchesVisibleWindow else { return }
                     window.touchesVisible = !window.touchesVisible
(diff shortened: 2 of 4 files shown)


Mastodon/Extension/UIApplication.swift
```diff
@@ -22,5 +22,13 @@ extension UIApplication {
...
         return version == build ? "v\(version)" : "v\(version) (\(build))"
     }
+    
+    func getKeyWindow() -> UIWindow? {
+        return UIApplication
+            .shared
+            .connectedScenes
+            .flatMap { ($0 as? UIWindowScene)?.windows ?? [] }
+            .first { $0.isKeyWindow }
+    }
```
cdoncarroll, line 32: Is there a better way to add extensions to UIKit classes in this project? Another extension in this same file with more targeted functionality? Another entire file even?

MastodonSDK/Sources/MastodonUI/View/Window/TouchesVisibleWindow.swift
```diff
@@ -0,0 +1,136 @@
+//
+//  File.swift
```
cdoncarroll: LOL. Will fix up.

MastodonSDK/Sources/MastodonUI/View/Window/TouchesVisibleWindow.swift
```diff
@@ -0,0 +1,136 @@
...
+                case .began:
+                    let touchView = newTouchView()
+                    touchView.center = touchLocation
+                    addSubview(touchView)
+                    touchViews[touch] = touchView
+                    
+                case .moved:
+                    guard let touchView = touchViews[touch] else {
+                        return
+                    }
```
tejuamirthi: just curious, do we not require to call `super.sendEvent(event)` here?

cdoncarroll: Good catch! There does need to be super call in this case. 

On second look, this early return actually feels kinda odd. I'll update this conditional to be an if-let, and we'll naturally fall through to the super call at the end of the function.

Mastodon/Extension/UIApplication.swift
zeitschlag: That's a good question, I'd say: It depends. At the moment, it's a bit inconsistent and WIP.

On one hand, there's an entire [folder full of extension-files](https://github.com/mastodon/mastodon-ios/tree/develop/Mastodon/Extension), on the other hand, the extensions are sprinkled around [here](https://github.com/mastodon/mastodon-ios/blob/5a5f93ad83c42b5bd7616909c56ec96895a713f5/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Attachment/AttachmentView.swift#L181) and [there](https://github.com/mastodon/mastodon-ios/blob/56efe8a93a3064139c347b43253d0c8a45f334b7/Mastodon/Scene/HashtagTimeline/HashtagTimelineHeaderView.swift#L15).

Personally, I'm a fan of putting them in their own file but I also see the elegance of a `fileprivate extension CGFloat`, if it only affects one file.

Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DebugAction.swift
```diff
@@ -122,6 +122,10 @@ extension HomeTimelineViewController {
             identifier: nil,
             options: [],
             children: [
+                UIAction(title: "Toggle Visible Touches", image: UIImage(systemName: "hand.tap"), attributes: []) { _ in
+                    guard let window = UIApplication.shared.getKeyWindow() as? TouchesVisibleWindow else { return }
+                    window.touchesVisible = !window.touchesVisible
```
zeitschlag, line 127: Why no [`toggle()`](https://developer.apple.com/documentation/swift/bool/toggle())?

zeitschlag:
Basically, this is only for people/developers who run the app using Xcode, right? I like it, it's a neat little thing, but I'm not so sure if adding more complexity is worth the benefit :/

Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DebugAction.swift
cdoncarroll: Long term Objective-C habits die hard I guess. 😅

cdoncarroll:
@zeitschlag, yes. It's a developer tool (just like any of the others in the debug menu). I've found in past projects that a similar option is really helpful in capturing more nuance when pitching a feature or reporting a bug. I actually already used it in this [pull request](https://github.com/mastodon/mastodon-ios/pull/746) to to make clearer the difference between the gesture and the result.

zeitschlag:
Alrighty, I'm convinced. What do you think @kimar @MainasuK ?

PR #741, opened 2022-12-07
[Feature] Translate Posts

kimar:
# Rationale

Translate post from it's original language to the appropriate one defined by the server.

# Tasks

- [x] Hookup with `/translate` endpoint
- [x] Implement translate action
- [x] Implement footer
- [x] Errorhandling if translation failed or is not available on the instance

# Demo

| Light mode | Dark mode |
|---|---|
| RocketSim_Recording_iPhone_14_Pro_2022-12-09_16 16 30| RocketSim_Recording_iPhone_14_Pro_2022-12-09_16 17 10|


MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Translation.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  File.swift
```
j-f1:
```suggestion
//  Mastodon+Entity+Translation.swift
```

zeitschlag (commented):
The code looks good, the feature is fun, too, but there might still be a layout issue somewhere:

I just wanted to translate [this one-line-post](https://mastodon.social/@Gargron/109490975321776720) and this is, what happened: 




No "Grüße der Saison", as it was visible on the profile afterwards:


 
Another example is this:





The last line just seems to ... get eaten? I'd suspect constraints or `self?.invalidateIntrinsicContentSize()` to cause this, but I'm pretty sure: You know better! 💪

j-f1:
It feels a bit weird to me that the banner appears below any media, especially since the media will not be translated.

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Translation.swift
DK1DA: Ha! Thanks @j-f1 !

kimar:
@j-f1 hm yeah this somewhat makes sense. Let me check and get back to this topic.

zeitschlag:
Another thing I stumbled upon: When I'm on the detail-page of a toot (it's the screenshot on the left), there's no option to translate it. Is this on purpose?

Mastodon/Protocol/Provider/DataSourceFacade+Translate.swift
```diff
@@ -0,0 +1,65 @@
...
+        let selectionFeedbackGenerator = await UISelectionFeedbackGenerator()
+        await selectionFeedbackGenerator.selectionChanged()
+
+        guard
+            let status = status.object(in: provider.context.managedObjectContext)
+        else {
+            return
+        }
+        
+        func translate(status: Status) async throws -> String? {
```
zeitschlag: I haven't seen this "a function in a function" very often yet, but I can image that it makes sense from to time.

Totally regardless of this and more out of curiosity: How would you write unit tests for this? Wouldn't this be easier by extracting this into their own functions?

MastodonSDK/Sources/MastodonCore/Extension/CoreDataStack/Instance.swift
```diff
@@ -24,9 +24,43 @@ extension Instance {
     }
 }
 
+extension Instance {
+    public var configurationV2: Mastodon.Entity.V2.Instance.Configuration? {
+        guard let configurationRaw = configurationV2Raw else { return nil }
+        guard let configuration = try? JSONDecoder().decode(Mastodon.Entity.V2.Instance.Configuration.self, from: configurationRaw) else {
+            return nil
+        }
```
zeitschlag: Maybe it makes sense to combine these two `guard`, like this:

```swift
guard let configurationRaw = configurationV2Raw, // or even guard let configurationV2Raw
    let configuration = try? ...
```

zeitschlag:
Regarding [this](https://github.com/mastodon/mastodon-ios/pull/741#pullrequestreview-1212620633): The last line appears now \o/ But the change is visible. There's a little flickering, With slow animations enabled it looks like this:

https://user-images.githubusercontent.com/2580019/207172232-df59b5fa-0637-4235-976d-e3c8ef39523f.mov

MastodonSDK/Sources/MastodonCore/Extension/CoreDataStack/Instance.swift
```diff
@@ -24,9 +24,43 @@ extension Instance {
...
+        return majorVersionInt >= comparedVersion
+    }
+}
+
+extension Instance {
+    var isTranslationEnabled: Bool {
+        if let configuration = configurationV2 {
+            return configuration.translation?.enabled == true
+        }
+        return false
```
zeitschlag: Maybe it makes sense to group `canFollowTags` and `isTranslationsEnabled` into one extension?

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+InstanceV2.swift
```diff
@@ -0,0 +1,158 @@
...
+//        public let maxCharacters: Int
+//        public let maxMediaAttachments: Int
+//        public let charactersReservedPerURL: Int
+//
+//        enum CodingKeys: String, CodingKey {
+//            case maxCharacters = "max_characters"
+//            case maxMediaAttachments = "max_media_attachments"
+//            case charactersReservedPerURL = "characters_reserved_per_url"
+//        }
+//    }
```
zeitschlag: What's the reason you want to keep this?

zeitschlag (commented):
Apart from the visible state-change (I leave that up to you :) ), it's good to go. I also like, that the translation-feature respects CWs 👍 

Screenshot 2022-12-13 at 00 07 41

Another thing is, that I get the option to translate German toots to ... German. Can you prevent this somehow?

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+InstanceV2.swift
DK1DA: Whoopsie! Sorry I completely forgot to remove this.

kimar:
@zeitschlag thanks a bunch for the thorough review! Yes I'll take a look at hiding the Translate option of source and target language are the same, that makes total sense.

Mastodon/Protocol/Provider/DataSourceFacade+Translate.swift
DK1DA: I like to use such nested functions whenever the functions will only be used by their parents and not anywhere else, and yes they are not testable in isolation but only by invoking the parent function. I'll extract them and make them private to have it more consistent with the rest of the code.

kimar:
@zeitschlag regarding the missing translate option on the detail page, there was a bug where the `AppContext` was not injected into the viewModel, thus not being able to determine if translations are enabled on the instance. This is fixed now, I'M not really a 100% happy with how this is injected into the view model, as it's optional there but I think fixing this will exceed the scope of this PR. I'll write it down to look at this later.

MainasuK:
> there was a bug where the AppContext was not injected into the viewModel…

That is what the #477 PR doing. The `AppContext` before this PR is a class in the main app target. And I'm glad to see you inject the missing context. 👍

MastodonSDK/Sources/MastodonCore/Service/API/CoreData/APIService+CoreData+InstanceV2.swift
```diff
@@ -0,0 +1,69 @@
+import os.log
+import Foundation
+import CoreData
+import CoreDataStack
+import MastodonSDK
+
+extension APIService.CoreData {
+    
+    static func createOrMergeInstanceV2(
```
MainasuK: I prefer using the same pattern in the `./MastodonSDK/Sources/MastodonCore/Persistence/Persistence+MastodonUser.swift` file. The function is same but maybe better naming style.

MastodonSDK/Sources/CoreDataStack/Entity/Mastodon/Status.swift
```diff
@@ -99,6 +99,8 @@ public final class Status: NSManagedObject {
     @NSManaged public private(set) var deletedAt: Date?
     // sourcery: autoUpdatableObject
     @NSManaged public private(set) var revealedAt: Date?
+    
+    @Published public var translatedContent: String?
```
MainasuK: Emmmmm. I'm not sure it is a good idea. If you want a property bound to an object. Please check Core Data `transient` property. The `Status.isSensitiveToggled` is an example. 

Also, the transient property lives in the same managed object context. So you always want to update and save in the `main` context only.

MainasuK: Let me explain. The Core Data object is not magic. The different core data object maybe represent the same `entity` (the OOP concept here is a matter).  There is no guarantee for reused cells using the previous object. Also, maybe the same object will be used between different timelines. e.g. home and thread. 

Using `transient` property on the object will sync the state between different locations. It's will be a help.

DK1DA: You're right! This totally makes sense. I'll fix this.

MastodonSDK/Sources/MastodonCore/Service/API/CoreData/APIService+CoreData+InstanceV2.swift
DK1DA: I'm not sure I understand this, you mean I should move the persistency code to something like `Persistence+Instance` ? Or is it something else you're talking about here?

MastodonSDK/Sources/CoreDataStack/Entity/Mastodon/Status.swift
DK1DA: Thank you @MainasuK done with https://github.com/mastodon/mastodon-ios/pull/741/commits/855f626c42fbb843090fc821a1b41830c6b63a9b

MastodonSDK/Sources/MastodonCore/Service/API/CoreData/APIService+CoreData+InstanceV2.swift
DK1DA: Ah I think now I got it, regarding the `PersistContext`, I'll update it to be similar.

Issue #752, opened 2022-12-13
[IMPROVEMENT] German translation for "Following" label in Profile view is weird


d4l-w4r:
Currently the label for the count of users a profile is following is translated as "Folge ich", which literally means "I follow". 
For the current users own profile page this is fine, but the same string is also used for other profiles, which looks a bit weird to me. 
For other profiles the correct string should read "folgt", which literally translates to "follows".

I feel like there's 3 Options:
1. Just don't translate and use the english "Following", as the "Follower" label is also not translated.
2. Use the non-personal form "Folgend"
3. Make a quick check for isSelf when the profile page loads and display different labels depending on whether it is the users own profile or another profile. 

Option 3 would be my preferred option, because it would result in the most clear and readable label (in my opinion).
Option 2 would feel like a regression, because that's how it used to be a while ago and it was horrible readability wise. Because Following was "folgend" and Follower was "folgende" and "following" can be translated to both folgend and folgende depending on context I always stumbled and wasn't sure what list I was looking at :D

If you think this is worthwhile and I don't receive pushback from fellow german speakers here, I will open another issue on the mastodon-android project and provide a PR for the implementation of Option 3 for Android. 
I can also try to provide the code for iOS.

The only thing I would need help with is the exact process of how to add a new key to the strings file and correctly get it into Crowdin.

In the end I think this could be an improvement for other languages as well where the current non-personal form is not as clear as the english "Following"/"Follower".

j-f1:
Check out https://github.com/mastodon/mastodon-ios/pull/753

d4l-w4r:
Oh cool! :)
Hang on I'll check if you caught all occurences...
I'm implementing this for Android right now and found 2 places where it needs to be changed. 
1. On the profile page
<img width="288" alt="Bildschirmfoto 2022-12-13 um 16 31 47" src="https://user-images.githubusercontent.com/4370186/207376074-ee38585d-ee56-4f59-9485-6ccceaadd90f.png">

2. In the header of the list of followers
<img width="288" alt="Bildschirmfoto 2022-12-13 um 16 32 32" src="https://user-images.githubusercontent.com/4370186/207376083-334c3c92-9b91-4fee-96cf-08dbddb0d0e4.png">

j-f1:
I’ve currently updated:

- profile pages (your first screenshot)
- profile cards (search → For You)

[...]

PR #753, opened 2022-12-13
Differentiate between “My followers” and other accounts followers for i18n

j-f1:
- See: https://github.com/mastodon/mastodon-ios/issues/752


d4l-w4r:
From what I can see I think this fixes the problem in the profile view.
It is missing one occurence though, when you click on the number of "following" on a profile, the label also shows up in the header of the resulting ListView:

Bildschirmfoto 2022-12-13 um 16 32 32

d4l-w4r (approved):
This looks like it solves my issue :)
I think including the distinction for my_followers and my_posts was a good idea as well!

PR #761, opened 2022-12-13
Use relative date formatting for long post timestamps

### LINKED ISSUE #759: Relative dates aren’t used in iOS
### Steps to reproduce the problem

Look at any timestamp on a post from today or yesterday.

### Expected behaviour

Today at 11:55 AM

### Actual behaviour

12/13/22, 11:55 AM

### Detailed description

It is best practice on iOS to use relative dates when formatting timestamps. This is a simple option on the DateFormatter. This produces (and localizes) “Today” and “Yesterday” prefixes when appropriate.

### Specifications

Latest iOS

j-f1:
This fixes #759.

That said, I don’t really like this area of the code. The `UIView.isZoomedMode || metricButtonTitleLength > [magic number]` check is weird to me and it would be great to rewrite it in SwiftUI to use the `ViewThatFits` view in iOS 16, or an equivalent UIKit construct that allows us to change the text of the label as the available space changes.

It’s also weird that the long and short forms of the date string are “Today at XX:XX” and “Today, XX:XX” in English (and there’s no good way to check for this case since the set of days for which relative formatting is applied depends on the locale)

Also not sure if I like this better than the current behavior but wanted to try it out

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
@@ -543,13 +543,14 @@ extension StatusView.ViewModel {
                 let formatter = DateFormatter()
                 
                 // make adaptive UI
-                if UIView.isZoomedMode || metricButtonTitleLength > 20 {
+                if UIView.isZoomedMode || metricButtonTitleLength > 30 {
                     formatter.dateStyle = .short
                     formatter.timeStyle = .short
                 } else {
                     formatter.dateStyle = .medium
                     formatter.timeStyle = .short
                 }
+                formatter.doesRelativeDateFormatting = true
                 return formatter.string(from: timestamp)
             }()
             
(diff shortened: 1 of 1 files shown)


Gargron:
> Also not sure if I like this better than the current behavior but wanted to try it out

I like the idea of "Today at xx:xx", personally

zeitschlag:
@j-f1 Thanks for all the work you're putting in :)

In this case I agree with @Gargron, so I'd suggest to close it, if that's fine for you?

j-f1:
The current behavior is to always show the full date (even for posts from yesterday/today), though. I could try having multiple labels and check their `intrinsicContentSize` or similar in `layoutSubviews` to only show the shorter versions if the longer one is too long.

I'm thinking:
- current long formatting
- current short formatting
- format with just the date (if >24h old) or just the time

j-f1:
Closing this since @zeitschlag will have the opportunity to make a similar change (and take advantage of increased horizontal real estate) due to the design changes in #875.

PR #771, opened 2022-12-15
Fix VoiceOver trap in thread titles

### LINKED ISSUE #542: [BUG] Accessibility/VoiceOver: Navigation trap at the top of a post detail view

### Current Behavior

- Go to your timeline view
- Turn on VoiceOver
- Swipe right to select a post
- Double tap to open its detail view
- VoiceOver focus will be at the back button √
- Swipe right 
- VoiceOver focus will be at the "Post from..." title √
- Swipe right
- Voiceover focus doesn't jump to the reply button. You'll just hear a "disabled" sound
- NOTE: This seems to be only happening when opening someone else' post, not posts of yourself

IMG_028C7CE59460-1

### Expected Behavior

I expect the VoiceOver focus to seamlessly move through the UI by swiping right on a detail view/post detail 

### Steps To Reproduce

See above

### Environment

```markdown
- Device: iPhone 12 mini
- OS: iOS 16.1
- Version: 1.4.5
- Build: 144
```

### Anything else?

_No response_

j-f1:
Fixes #542! I’m not sure why this fixes the issue but it does. Now the `DoubleTitleLabelNavigationBarTitleView` sets itself as an accessibility element with no children, and sets its label to `title, subtitle`. Navigation through the title area now works properly!

### DIFF
--- a/Mastodon/Scene/Share/View/Content/DoubleTitleLabelNavigationBarTitleView.swift
+++ b/Mastodon/Scene/Share/View/Content/DoubleTitleLabelNavigationBarTitleView.swift
@@ -51,16 +51,20 @@ extension DoubleTitleLabelNavigationBarTitleView {
         
         containerView.addArrangedSubview(titleLabel)
         containerView.addArrangedSubview(subtitleLabel)
+
+        isAccessibilityElement = true
     }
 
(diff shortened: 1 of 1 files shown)


kimar (approved):
LGTM, thanks you!

PR #772, opened 2022-12-15
Add a drag interaction and link preview to post links and cards

### LINKED ISSUE #652: [BUG] Standard tap and hold link action cannot be used on posts

### Current Behavior

Tap and hold link (to copy URL without opening page) doesn’t perform the expected action.

Instead it either opens the link – when in main timeline view – or selects text in discreet post view, which is not useful as link is normally elided (blah.com/theconten…).

### Expected Behavior

Holding the link in either context (timeline/post) should invoke the standard iOS behaviour and pop up the link sharing/preview actions.

A06C9856-FE86-4AC7-9B5F-2B806A139AAE

### Steps To Reproduce

View any post with link; tap and hold.

### Environment

```markdown
- Device: iPhone 12
- OS: iOS
- Version: 16.1
- Build:
```

### Anything else?

_No response_

### LINKED ISSUE #725: [FEATURE] Make it possible to copy a link
It would be helpful, if it would be possible to copy a link which is included in a message. Currently the only way to do this is to tap on the link, open it in the browser and copy it there.

j-f1:
- Fixes #652
- Fixes #725
- add a drag interaction to post links and the card so you can drop the URL elsewhere (like into a new post window or something)
- add a context menu to links in posts with the same actions as on the card
- Configure an `SFSafariViewController` to enable link previews when long pressing links/cards
  - this works exactly like in Mail/Safari/other apps, including support for toggling the preview on and off!
  - (implemented) ~it would be ideal if there were some way to pass the `SFSafariViewController` instance up to the scene coordinator so that if the user commits the long press interaction the page can be shown immediately without having to reload it (and also ideally it would be a synchronous method call so that the animated presentation could be adjusted to occur in sync with the commit animation).~

code changes:
- unify `statusView:didTapCardWithURL:` and `statusView:cardControl:didTapURL:` since they do the same thing
- `SceneCoordinator` now allows you to fetch a view controller for a proposed scene separately from doing the transition, this allows for peek+pop to be added elsewhere (posts? other meta entities? user cards?)
- `UITextView` now has support for finding the `Meta` object corresponding to a point and snapshotting a text range (for use in the highlight preview)
- The code is extensible to other meta types (like hashtags and mentions), I may implement these in future PRs if I have time.

~This is still WIP because I would like to configure:~
- [x] a drag interaction on links too
- [x] some sort of code sharing arrangement since currently the code for the context menu stuff is a little duplicative (it’s better now but there could be improvements?)
- [x] some more code sharing to make it easier to enable previews/context menus/drag interactions for hashtags and mentions in the future (specifically in the code that creates the targeted preview for the drag/context menu interactions)

### DIFF
--- a/Mastodon/Activity/SafariActivity.swift
+++ b/Mastodon/Activity/SafariActivity.swift
@@ -6,7 +6,6 @@
 //
 
 import UIKit
-import SafariServices
 import MastodonAsset
 import MastodonLocalization
 
--- a/Mastodon/Coordinator/SceneCoordinator.swift
+++ b/Mastodon/Coordinator/SceneCoordinator.swift
@@ -275,8 +275,8 @@ extension SceneCoordinator {
 
     @MainActor
     @discardableResult
-    func present(scene: Scene, from sender: UIViewController? = nil, transition: Transition) -> UIViewController? {
-        guard let viewController = get(scene: scene) else {
+    func present(scene: Scene, from sender: UIViewController?, to prebuiltViewController: UIViewController? = nil, transition: Transition) -> UIViewController? {
+        guard let viewController = prebuiltViewController ?? viewController(forScene: scene) else {
             return nil
         }
         guard var presentingViewController = sender ?? sceneDelegate.window?.rootViewController?.topMost else {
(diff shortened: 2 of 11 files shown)


j-f1:
wow, this is harder than I expected. ~Current issue is that expanding a link preview and then switching to a drag results in a blank drag preview. ¯\\\_(ツ)\_/¯~ Reported as FB11957656

~Also need to figure out how to wire the context menu items up to the link drag.~ done

j-f1:
Ok, I think I’ve finally hammered out my qualms with this PR (except for the bug that I’ve filed with Apple above). Let me know if I should split this up or if you’re OK with reviewing it as-is!

j-f1:
Not sure why the checks are failing, maybe some sort of build caching issue? works fine on my end.

PR #773, opened 2022-12-15
Update to MetaTextKit 4

j-f1:
This brings the app up to the latest version of MetaTextKit and drops support for iOS 14, allowing the use of more modern APIs.

- @MainasuK can you take a look a the changes in this PR to make sure I’m using the updated API correctly? Especially the changes in `MastodonUI/Extension/MetaLabel.swift`.
- I couldn’t spot any issues with the new version after doing a quick round of testing but this should definitely be run for a while on beta/dev devices to make sure nothing has broken


zeitschlag:
@j-f1 Hey, forgive me for changing the target-branch of this PR 🙈 What bugged me was that this PR changed the deployment target, but didn't change any of the `if #available(iOS 15.0, *)`, so I just made my own PR to fix that.

So, basically, all, that this PR changes now is updating MetaTextKit to version 4 (that's at least what it should do, you probably know way better than me). I hope, that this is okay for you?

j-f1:
No worries, and it makes sense to split the changes into two PRs anyway IMO!

MainasuK:
I'm sorry for the late reply to this issue. The `MetaTextKit` is a framework shared by multiple projects which I maintain. The goal of this framework is to provide the Label and TextView with link and view attachments (e.g. APNG emojis) supports.

The v2 is the latest version before the framework transformed to the iOS new `TextKit2` framework. But I have to say the `TextKit2` is not ready in iOS 15. The RTL supports is much worse than the `TextKit 1` (even the same issue in iOS 16). Maybe we just leave it here because the old one works now.

I do not recommend upgrading that, at least, before the RTL layout problem is resolved.

MainasuK:
That is a v4 version demo on Twidere X (v1.4.3). The RTL layout somehow overflows to the avatar. But the LTR is no problem.


| RTL  | LTR  |
| ------------- | ------------- |
|   |   |

j-f1:
No worries, thanks for the info about the bugs in TextKit 2! I think it makes sense to close this PR for now, then.

~I’m still interested in getting the changes from https://github.com/TwidereProject/MetaTextKit/pull/2 into the Mastodon app so the VoiceOver experience can be better, so I’ll retarget that PR to the `v2` branch. Would that make sense?~ never mind, this is better done inside the app.

PR #775, opened 2022-12-15
Add a CI job that blocks PRs from modifying old Core Data models

j-f1:
This check helps prevent at least two kinds of issues:

1. Accidentally changing old models (which causes could cause issues with migrating people running old app versions)
2. Changing the current version of the model, then having your PR not be merged while new model versions are created, then doing a merge without realizing that your commits no longer update the current version of the model.

See the checks below for an example of the error message 🤞


sladewatkins (approved):

j-f1:
@zeitschlag why the 👎?

PR #783, opened 2022-12-17
Bump Deplyment Target to iOS 15

zeitschlag:


Mastodon/Extension/UICollectionViewDiffableDataSource.swift
```diff
@@ -13,28 +13,14 @@ extension UICollectionViewDiffableDataSource {
...
     ) {
-        if #available(iOS 15.0, *) {
-            self.applySnapshotUsingReloadData(snapshot, completion: completion)
-        } else {
-            self.apply(snapshot, animatingDifferences: false, completion: completion)
-        }
+        self.applySnapshotUsingReloadData(snapshot, completion: completion)
     }
     
     func applySnapshot(
```
j-f1: I think these functions can be dropped entirely in favor of just calling `apply(snapshot, ...)`, right?

MastodonSDK/Sources/MastodonUI/Vendor/MetaTextView+PasteExtensions.swift
```diff
@@ -15,19 +15,17 @@ extension MetaTextView {
...
-            var nextResponder = self.next;
-            
-            // Force the event to bubble through ALL responders
-            // This is a workaround as somewhere down the chain the paste event gets eaten
-            while (nextResponder != nil) {
-                if let nextResponder = nextResponder {
-                    if (nextResponder.responds(to: #selector(UIResponderStandardEditActions.paste(_:)))) {
-                        nextResponder.perform(#selector(UIResponderStandardEditActions.paste(_:)), with: sender)
-                    }
+        var nextResponder = self.next;
```
j-f1, line 18: This code could be rewritten using [`sequence(first:next:)`](https://developer.apple.com/documentation/swift/sequence(first:next:)) (although I’m not sure if it needs to be)

```swift
for responder in sequence(first: self, next: \.next) {
    if responder.responds(to: #selector(UIResponderStandardEditActions.paste(_:))) {
        responder.perform(#selector(UIResponderStandardEditActions.paste(_:)), with: sender)
    }
}
```

j-f1:
Also, are you working on handling the deprecation warnings? I can take a look at those if you’re not!

Mastodon/Extension/UICollectionViewDiffableDataSource.swift
zeitschlag: You are absolutely right 👍 
Great, that you saw that :)

zeitschlag:
@j-f1 Good point. If you see something, feel free to fix it :) I'll do so myself, too. Thanks for you help!

From what I see, most warnings are `'contentEdgeInsets' was deprecated in iOS 15.0: This property is ignored when using UIButtonConfiguration`, but we don't use configurations (yet)

j-f1 (approved):
LGTM!

sladewatkins (approved):
LGTM as well!

kimar (approved):
LGTM!

Issue #309, opened 2021-12-03
Voice over bugs


Gargron:
Analysis provided by @marcozehe [here](https://toot.cafe/@marcozehe/107352653997388458)

- [ ] The text in the header says „mastodon.text“, claims is a button, but double-tapping only scrolls to the top
- [ ] Posts behind CWs aren‘t indicated. I get the content without warning.
- [ ] Time stamps spoken get terribly wrong once I start scrolling the screen
- [ ] Notification screen has multiple stops per item
- [ ] Posts in main screen have no custom actions to reply etc
- [ ] Poll options and their current values in my original post aren‘t accessible at all.
- [ ] The theme of messed up time stamps on posts continues in detail/conversation screens.
- [ ] Reply button claims there are 0 reblogs if a post has no replies, boosts and favorites. Even if it has, the Reply button still talks about reblogs instead of replies.

Bracciata:
In the current version, the header's text says " logo button button". Do you have an opinion on what it should be upgraded to? Potentially "Scroll to top button"?

PR #784, opened 2022-12-18
fix: updated logo button's alt text to be more accessible

Bracciata:
This covers "The text in the header says „mastodon.text“, claims is a button, but double-tapping only scrolls to the top" from issue #309.

This update adds the necessary information so that users are aware of what the button does. In addition, the word button in the text is redundant since it is a button so it pops up again.


zeitschlag:
@Bracciata Hey, thank you for your work 🥳 You address a very fair and important point.

But: Most of your changes will be overriden when we update the localization-files. As described [here][localization-readme], we use [crowdin][crowdin] for translations. So: The translations need to be updated on crowdin :/

Also there might be a little misunderstanding: Tapping the Mastodon-button (or: the entire NavigationBar) at the top scrolls to the top, yes, but it also scrolls back to the position you were at before you tapped the Mastodon-button. Do you know, what I mean?

[localization-readme]: https://github.com/mastodon/mastodon-ios/blob/develop/Localization/README.md
[crowdin]: https://crowdin.com/project/mastodon-for-ios

Bracciata:
@zeitschlag thank you for the review. I missed the accessibility hint earlier. I reupdated the alt text to remove the additional button and also made it the desired message of what the text within the image is ("Mastodon").

zeitschlag:
Hey @Bracciata,

when reviewing your PR, I found out: We do not even use those keys 🙈 So: Thanks for pointing that out by opening the PR. Do you feel confident to implement the needed change? If you need any help: Just say something. This is the code that needs to be changed:

https://github.com/mastodon/mastodon-ios/blob/d4eabf95d17a5145f98345d210891eec6a2a9109/Mastodon/Scene/HomeTimeline/View/HomeTimelineNavigationBarTitleView.swift#L106-L107

Bracciata:
@zeitschlag I made the fix, would you rather I update the code to use the keys?

zeitschlag:
@Bracciata Yeah, that was the idea: To give you the opportunity to contribute to the project yourself 🙂

Bracciata:
@zeitschlag Should be good to go! Let me know if any other changes are needed as I am new to iOS development. Thank you for your patience.

PR #817, opened 2022-12-24
Improve scrolling performance in threads

### LINKED ISSUE #150: Profile the timeline scrolling performance
Needs to figure out the most time cost task and make the scrolling more smooth. Especially the home scene.

j-f1:
I profiled using Instruments and the largest cause of hitches appeared to be the synchronous call to `status.managedObjectContext?.safeFetch` to get the info for the user being replied to. I was able to move this to a background thread. I also took the opportunity to split out the functionality here to reduce nesting and hopefully make it easier to understand for folks making updates into the future.

The one negative consequence is that now when you scroll rapidly through a thread you’ll briefly see “Replied to …” for a fraction of a second before the username is loaded.

probably closes #150 (although that issue is very old, I wasn’t able to find any other scrolling hangs on my 13 Pro)

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Extension/Task.swift
+++ b/MastodonSDK/Sources/MastodonUI/Extension/Task.swift
@@ -1,0 +1,14 @@
+//
+//  Task.swift
+//  
+//
+//  Created by Jed Fox on 2022-12-23.
+//
+
+import Combine
+
+extension Task {
+    func store(in set: inout Set<AnyCancellable>) {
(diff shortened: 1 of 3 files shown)


j-f1:
~…wait a minute. How are `MastodonUser`s keyed? Is the `domain` property supposed to be the domain they were fetched from or is it the domain the actual account was created on?~

j-f1:
Ok, I think this should be reasonably correct now.

crimepoppin (approved):

Issue #664, opened 2022-11-21
[BUG] Saving video does nothing


jmeosbn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Tap and hold on a video offers to “Save Photo”, “Copy Photo”, or “Share”.

The first two options fail to do anything; the last provides the URL to the video (only).

### Expected Behavior

The first option should save the video to the iOS photo’s library, the second should probably just copy the URL (and they should be labelled as “Save/Copy Video”).

The third can share both the URL and the video file (system provides one depending on context; so user can choose, eg. to copy link, or save file to Files app).

### Steps To Reproduce

Tap and hold on a video embedded in a post.

Choose from the above options.

### Environment

```markdown
- Device: iPhone 12
- OS: iOS
- Version: 16.1
- Build:
```

PR #818, opened 2022-12-24
Implement Save Video. 

jinsu35:
#664 When a user taps on medium, he can either save it to Photos Album or share it.

`PHAssetCreationRequest.addResource(with:data:options:)` does not work with video for a reason unknown. The only way around it is to use `addResource(with:fileURL:options:)` instead which requires writing data to a temporary file first.


MastodonSDK/Sources/MastodonCore/Service/PhotoLibraryService.swift
```diff
@@ -130,23 +130,52 @@ extension PhotoLibraryService {
...
-        return Future<Void, Error> { promise in
-            PHPhotoLibrary.shared().performChanges {
-                PHAssetCreationRequest.forAsset().addResource(with: .photo, data: imageData, options: nil)
-            } completionHandler: { isSuccess, error in
-                if let error = error {
-                    promise(.failure(error))
-                } else {
-                    promise(.success(Void()))
+        
+        if imageType == .video {
```
j-f1: Would it make sense for this to be a `switch` so that all possible asset types are either explicitly handled or have an assertion raised?

MastodonSDK/Sources/MastodonCore/Service/PhotoLibraryService.swift
```diff
@@ -130,23 +130,52 @@ extension PhotoLibraryService {
...
+            return Future<Void, Error> { promise in
+                let filename = UUID().uuidString
+                let path = FileManager.default.temporaryDirectory.appendingPathComponent("\(filename)")
+                let url = path.appendingPathExtension("mp4")
```
j-f1: Is it a problem that videos might not be MP4 formatted (eg there’s HEVC and WebM and AV1 which instances might support to reduce disk usage and improve video quality)?

Mastodon/Protocol/Provider/DataSourceProvider+UITableViewDelegate.swift
```diff
@@ -166,30 +175,39 @@ extension UITableViewDelegate where Self: DataSourceProvider & MediaPreviewableV
...
+                        }
+                    }
+                )
+                return UIMenu(
+                    title: "",
+                    image: nil,
+                    identifier: nil,
+                    options: [],
+                    children: subMenu
                 )
```
j-f1: Small nit: this should work too:
```suggestion
                return UIMenu(children: subMenu)
```

Mastodon/Protocol/Provider/DataSourceProvider+UITableViewDelegate.swift
```diff
@@ -166,30 +175,39 @@ extension UITableViewDelegate where Self: DataSourceProvider & MediaPreviewableV
...
+                    )
+                }
+                subMenu.append(
+                    UIAction(
+                        title: L10n.Common.Controls.Actions.share,
+                        image: UIImage(systemName: "square.and.arrow.up")!,
+                        identifier: nil,
+                        discoverabilityTitle: nil,
+                        attributes: [],
+                        state: .off
```
j-f1: I believe most of these parameters can be removed since they have default values (applies to the other `UIAction` constructors too)

MastodonSDK/Sources/MastodonCore/Service/PhotoLibraryService.swift
```diff
@@ -32,7 +32,7 @@ extension PhotoLibraryService {
 
 extension PhotoLibraryService {
 
-    public func save(imageSource source: ImageSource) -> AnyPublisher<Void, Error> {
+    public func save(imageSource source: ImageSource, imageType: PHAssetResourceType = .photo) -> AnyPublisher<Void, Error> {
```
j-f1: Maybe this could use a rename now that it handles video too?

MastodonSDK/Sources/MastodonCore/Service/PhotoLibraryService.swift
```diff
@@ -130,23 +130,52 @@ extension PhotoLibraryService {
...
+        if imageType == .video {
+            return Future<Void, Error> { promise in
+                let filename = UUID().uuidString
+                let path = FileManager.default.temporaryDirectory.appendingPathComponent("\(filename)")
+                let url = path.appendingPathExtension("mp4")
+                do {
+                    try imageData.write(to: url)
+                } catch {
+                    print(error)
+                }
```
j-f1: If the file can’t be written, the video will not be able to be added to the library and so this should fail the future immediately, right?

PR #820, opened 2022-12-25
Fix: Profile Avatar Border Color is correct in Dark Mode

pietropizzi:
I noticed that the border around the avatar was always white, not matter if the UI was in light or dark mode.

Now the border color is the same as the background color around it, the same as it was in light mode.

| Before | After |
| -------|------|
| SCR-20221225-nh3 | SCR-20221225-nfk |


zeitschlag:
Hey @pietropizzi first of all: merry christmas (if you celebrate), it's not too late yet! Aaaand, of course: Thank you for this PR. Looks good to me 👍

pietropizzi:
> Hey @pietropizzi first of all: merry christmas (if you celebrate), it's not too late yet! Aaaand, of course: Thank you for this PR. Looks good to me 👍 

Thank you for merging @zeitschlag. First PR here, so I'm happy it got accepted.

Grüße aus Berlin

Issue #463, opened 2022-07-03
[BUG] <Emoji keyboard counts characters differently from web app>


gobborg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

On the Mastodon iOS app, you can use Mastodon's built-in custom emojis OR you can use the emoji keyboard.

The Mastodon web app counts each emoji as 2 characters.
When using the iOS keyboard in the Mastodon app, each emoji counts as 1 character.

Custom emojis have varying character counts for both web app and on the iOS app.

### Expected Behavior

Emojis in both the web app and the iOS app would count for the same amount of characters. Presumably, this would be 2 characters, unless the web app changes how it counts characters.

### Steps To Reproduce

In the Mastodon web app, try using noncustom emojis and observe that the character counter decrements in 2s.
In the Mastodon iOS app, use the iOS emoji keyboard and observe that the character counter decrements in 1s.
![image](https://user-images.githubusercontent.com/81018718/177020105-0c119fb0-d378-4f76-bb27-fc375060818f.png)
![image](https://user-images.githubusercontent.com/81018718/177020111-6e0560e5-01e3-46da-b916-24ae3f63c818.png)

### Environment

```markdown
- Device: iPhone 13 mini
- OS: iOS 15.5
- Version: 1.4.5
- Build: ?
```

PR #822, opened 2022-12-26
Use twitter-text parsing libary to count characters of content

jinsu35:
#463


j-f1:
One thing that would be hard to take into account here is that Mastodon servers are allowed to change the equivalent length of URLs if desired. It also doesn’t handle that characters starting with the second `@` in `@foo@bar.baz` are not counted. I have another PR for a closely related issue (https://github.com/mastodon/mastodon-ios/issues/669) but that definitely needs to be updated to 1) use the constant from the server and 2) use the proper string encoding (is it UTF-16? Because that's what JS uses, but the default for Ruby seems to be UTF-8)

PR #825, opened 2022-12-27
Call apiService in viewDidAppear

### LINKED ISSUE #417: [BUG] Own profile's following count doesn't update

### Current Behavior

In the "My Profile" tab (where you see your own profile), the number of accounts I follow (following count) doesn't update. In my case, the displayed value is 3, but it should be 7.

Restarting the app or refreshing the view (by pulling down) both don't help.
The only thing that did help was to create a new post, reloading the timeline and then going back to the profile tab.

The number of posts as well as the follower count also seem to be affected. After I deleted the aforementioned post, the number of posts still shows 1 instead of 0.

### Expected Behavior

The number updates each time the tab is refreshed (by pull-to-refresh) or reloaded after restarting the app.

### Steps To Reproduce

1. Go to your profile tab
2. Look at the number of accounts you follow ("Following")
3. Remember that number
4. Search for an arbitrary user you don't follow yet and start following
5. Go back to the profile tab and pull to refresh
6. Observation: the following count didn't change

### Environment

```markdown
- Device: iPhone X
- OS: iOS 15.4.1
- Version: 1.4.0
- Build: 121
```

### Anything else?

_No response_

jinsu35:
Resolves #417
Call api services for followers and following in viewDidAppear() in ProfileHeaderViewController.

### DIFF
--- a/Mastodon/Scene/Profile/Header/ProfileHeaderViewController.swift
+++ b/Mastodon/Scene/Profile/Header/ProfileHeaderViewController.swift
@@ -183,6 +183,32 @@ extension ProfileHeaderViewController {
         
         // set display after view appear
         profileHeaderView.setupImageOverlayViews()
+        
+        guard let userID = viewModel.user?.id else { return }
+        Task {
+            do {
+                let response = try await viewModel.context.apiService.followers(
+                    userID: userID,
+                    maxID: nil,
+                    authenticationBox: viewModel.authContext.mastodonAuthenticationBox
+                )
+                
+                profileHeaderView.viewModel.followersCount = response.value.count
+            } catch {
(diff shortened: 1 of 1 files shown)


j-f1:
It looks like this doesn’t behave correctly if there are more than 40 results since it will only return a value up to 40. See: https://discord.com/channels/231908446830723072/289546924217073664/1112434439985823885

kimar:
Yes @j-f1 you're right, if it's just about the counts we could also use the account details right? Let me know if anyone wants to fix this, otherwise I can do it tomorrow.

j-f1:
I can open a PR, I think we should be fetching the entire account (https://docs.joinmastodon.org/methods/accounts/#get) here which includes follow/follower counts directly instead of having to paginate a potentially huge list.

PR #834, opened 2022-12-28
fix: Poll percentages were wrong due to wrong count being used (votes vs voters)

kimar:
# Rationale

Fix incorrect poll votes percentage.

Reference status with which to reproduce issue with current App Store version: https://mastodon.social/@Gargron/109502941346549786

# Demo

| Before | After |
|---|---|
| Bildschirm­foto 2022-12-28 um 12 42 57 | Bildschirm­foto 2022-12-28 um 12 39 35 |

### DIFF
--- a/Mastodon/Scene/Share/View/Content/PollOptionView+Configuration.swift
+++ b/Mastodon/Scene/Share/View/Content/PollOptionView+Configuration.swift
@@ -33,12 +33,12 @@ extension PollOptionView {
             .store(in: &disposeBag)
         // percentage
         Publishers.CombineLatest(
-            option.poll.publisher(for: \.votesCount),
+            option.poll.publisher(for: \.votersCount),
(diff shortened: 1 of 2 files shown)


MainasuK:
The sum for all poll's options percentage in `After` is: 53% + 22% + 45% + 47% = 167%. That's more than 100%. Maybe something is wrong? 🤔 

I'm not researching the multiple polls on the webpage. So the same result should be correct. And that surprises me.

MainasuK (approved):

Gargron:
> The sum for all poll's options percentage in `After` is: 53% + 22% + 45% + 47% = 167%. That's more than 100%. Maybe something is wrong? 🤔

There is no point summing up percentages of a multiple choice poll.

zeitschlag (approved):

PR #836, opened 2022-12-28
Limit posts to 15 lines unless they are expanded

### LINKED ISSUE #730: Option to truncate long toots until expanded
I would love an option to truncate long toots until clicked on to expand to see the full text.  If this could be a user configurable number of words or characters that would be ideal.

j-f1:
Fixes #730. I went with ~20~ 15 lines since it felt long enough to accommodate most content but not allow things to get too much longer than one screen. When you tap on a post you’ll be always be able to see its full content.

Currently the only indication of whether a post is truncated is the “…” at the end. I considered doing a “Read More” button (inline or on a new line) but it seems that doing so is a bit messy since it requires iterating through all of the lines in the post until a truncation is found. (and whether or not the post is truncated could change dynamically based on text size or (on iPad) multitasking status).

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/StatusView.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/StatusView.swift
@@ -447,6 +447,8 @@ extension StatusView.Style {
         statusView.contentContainer.setContentCompressionResistancePriority(.required - 1, for: .vertical)
 
         // status content
+        statusView.contentMetaText.textView.textContainer.maximumNumberOfLines = 20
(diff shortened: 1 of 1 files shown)


j-f1:
Hmmm maybe like 15 lines would be better? Considering that there are link cards and also larger font sizes have larger line heights, and smaller phones exist, the current limit might be taller than the screen.

j-f1:
New limit of 15 is now comfortably within the screen size on my 13 Pro non-Max at the default font size for text-only posts, but media posts can be taller than the screen. I think that’s fine.

PR #841, opened 2022-12-30
Fix bug - scrollToTop() should be called when current tab is selected

### LINKED ISSUE #744: [BUG] main feed tab forgets my place after visiting any other tab

### Current Behavior

I scroll down my main feed. When I switch to another tab (search or notifications or my profile) and then go back to my main feed, it has lost my place and gone back to the top.

### Expected Behavior

I would expect the main feed to remember where I was so that I can continue reading chronologically. This is especially annoying when I switch to another tab inadvertently (often switching to my profile by accidentally brushing my phones screen with the heel of my palm, for instance).

### Steps To Reproduce

Pretty much as described above.

1. scroll any distance through the main feed
2. Switch to any other tab
3. Return to main feed

### Environment

```markdown
- Device: iPhone 12
- OS: 15.6.1
- Version: 1.4.8
- Build: 219
```

### Anything else?

I am reasonably sure this behavior started with the most recent update. (The one that added hashtag-following)
(tried to attach screenshots but github borked them. Feel like it’s self-explanatory but if screenshots would help I’ll try again.)

jinsu35:
Resolves #744

### DIFF
--- a/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
+++ b/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
@@ -572,7 +572,7 @@ extension MainTabBarController: UITabBarControllerDelegate {
 
         // Assert index is as same as the tab rawValue. This check needs to be done `shouldSelect`
         // because the nav controller has already popped in `didSelect`.
-        if currentTab.rawValue == tabBarController.selectedIndex,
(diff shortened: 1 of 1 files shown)


j-f1:
@kimar Would it be possible to get this PR merged? It fixes a rather annoying bug that’s been around for a few months now.

PR #847, opened 2023-01-01
Implement URL scheme

### LINKED ISSUE #391: URL scheme for opening Mastodon links
Adding a URL scheme for opening web links to Mastodon entities would let me integrate it into [Opener](https://opener.link/get)!

jinsu35:
Resolves #391 

The URL scheme structure and its functions are as following.

1. `mastodon://profile/username@instance` opens a user's profile page.
2. `mastodon://status/:id` shows a status.
3. `mastodon://post` starts a new post.


Mastodon/Scene/Profile/RemoteProfileViewModel.swift
```diff
@@ -17,38 +17,42 @@ final class RemoteProfileViewModel: ProfileViewModel {
...
-            }
-            .retry(3)
-            .sink { completion in
-                switch completion {
-                case .failure(let error):
-                    // TODO: handle error
-                    os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s: remote user %s fetch failed: %s", ((#file as NSString).lastPathComponent), #line, #function, userID, error.localizedDescription)
-                case .finished:
-                    os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s: remote user %s fetched", ((#file as NSString).lastPathComponent), #line, #function, userID)
+        Task { @MainActor in
```
j-f1: Why this change? I don’t think it does anything to guarantee whee the `sink` will be called.

Mastodon/Supporting Files/SceneDelegate.swift
```diff
@@ -219,4 +210,89 @@ extension SceneDelegate {
...
+                logger.debug("\((#file as NSString).lastPathComponent, privacy: .public)[\(#line, privacy: .public)], \(#function, privacy: .public): not authenticated")
+            }
+        }
+    }
+    
+    private func handleUrl(context: UIOpenURLContext) {
+        let sendingAppID = context.options.sourceApplication
+        let url = context.url
+
+        if !UIApplication.shared.canOpenURL(url) { return }
```
j-f1, line 243: What does this do?

Mastodon/Supporting Files/SceneDelegate.swift
```diff
@@ -219,4 +210,89 @@ extension SceneDelegate {
...
+        if let username = url.user {
+            guard let host = url.host else { return }
+            let components = url.pathComponents
+            if components.count == 3 && components[1] == "status" {
+                let statusId = components[2]
+                // View post from user
+                print("view status \(statusId)")
+                if let authContext = coordinator?.authContext {
+                    Task {
+                        guard let thread = try await AppContext.shared.apiService.fetchThread(
```
j-f1: Given that fetching the thread can take some time, would it be possible to show the thread view immediately and then fetch the content in parallel?

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Thread.swift
```diff
@@ -16,9 +16,10 @@ extension APIService {
     
     public func statusContext(
         statusID: Mastodon.Entity.Status.ID,
-        authenticationBox: MastodonAuthenticationBox
+        authenticationBox: MastodonAuthenticationBox,
+        domain: String? = nil
     ) async throws -> Mastodon.Response.Content<Mastodon.Entity.Context> {
-        let domain = authenticationBox.domain
+        let domain = domain ?? authenticationBox.domain
```
j-f1: I don’t think this is correct — you always want to send requests to `authenticationBox.domain` or else you’ll be leaking the auth token to a different server. Since status IDs are unique (using the snowflake scheme) you should be able to request it from the currently authenticated server.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Thread.swift
```diff
@@ -51,4 +52,79 @@ extension APIService {
...
+        statusID: Mastodon.Entity.Status.ID,
+        domain: String,
+        authenticationBox: MastodonAuthenticationBox
+    ) async throws -> Status? {
+        let authorization = authenticationBox.userAuthorization
+        let managedObjectContext = self.backgroundManagedObjectContext
+
+        let responseOne = try await Mastodon.API.Statuses.status(
+            session: session,
+            domain: domain,
```
j-f1: same here wrt auth and looking up remote statuses

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Account.swift
```diff
@@ -199,3 +200,43 @@ extension APIService {
         return response
     }   // end func
 }
+
+extension APIService {
+    public func fetchUser(username: String, domain: String, authenticationBox: MastodonAuthenticationBox)
```
j-f1, line 205: This could be cool to use for handling taps on mentions in profile bios too! (definitely a separate PR tho)

j-f1 (commented):
This should be decided by someone other than me but IMO the scheme should be `joinmastodon://` so that other apps can detect the presence of specifically this app, and so that `mastodon://` can be standardized among apps across platforms later.

Mastodon/Supporting Files/SceneDelegate.swift
jinsu35: It checks the validity of the url.

Mastodon/Supporting Files/SceneDelegate.swift
jinsu35: Removed redundant fetching of data in fetchThread() which is taken care by ThreadViewModel so the function returns fast.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Thread.swift
```diff
@@ -51,4 +51,79 @@ extension APIService {
         return response
     }   // end func
     
+    public func fetchThread(
```
DK1DA: The `APIService` is supposed to return types of `Mastodon.Response.Content`.
Also I think this code needs to duplication but the actual logic is already existing, you should probably reuse the APIService to fetch a status and grab the persisted object on the call-site.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Account.swift
DK1DA, line 205: This function actually is not required, what you want to achieve can be done using modified initializer on `RemoteProfileViewModel` which, instead of querying by id, uses the user's full identifier to look them up.

In `RemoteProfileViewModel`:

```swift
    init(context: AppContext, authContext: AuthContext, acct: String) {
        super.init(context: context, authContext: authContext, optionalMastodonUser: nil)
        
        let domain = authContext.mastodonAuthenticationBox.domain
        let authorization = authContext.mastodonAuthenticationBox.userAuthorization
        Just(acct)
            .asyncMap { acct in
                try await context.apiService.accountLookup(
                    domain: domain,
                    query: .init(acct: acct),
                    authorization: authorization
                ).singleOutput()
            }
            .retry(3)
            .sink { completion in
                switch completion {
                case .failure(let error):
                    // TODO: handle error
                    os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s: remote user %s fetch failed: %s", ((#file as NSString).lastPathComponent), #line, #function, acct, error.localizedDescription)
                case .finished:
                    os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s: remote user %s fetched", ((#file as NSString).lastPathComponent), #line, #function, acct)
                }
            } receiveValue: { [weak self] response in
                let managedObjectContext = context.managedObjectContext
                let request = MastodonUser.sortedFetchRequest
                request.fetchLimit = 1
                request.predicate = MastodonUser.predicate(domain: domain, id: response.value.id)
                guard let mastodonUser = managedObjectContext.safeFetch(request).first else {
                    assertionFailure()
                    return
                }
                DispatchQueue.main.async {
                    self.user = mastodonUser
                }
            }
            .store(in: &disposeBag)
    }
```

Mastodon/Supporting Files/SceneDelegate.swift
```diff
@@ -219,4 +210,93 @@ extension SceneDelegate {
...
+            showComposeViewController()
+        case "profile":
+            let components = url.pathComponents
+            if components.count == 2 && components[0] == "/" {
+                let addr = components[1]
+                let tokens = addr.components(separatedBy: "@")
+                if tokens.count != 2 { return }
+                let username = tokens[0]
+                let host = tokens[1]
+                if let authContext = coordinator?.authContext {
```
DK1DA, line 255: When using the suggested initializer in `RemoteProfileViewModel` you could just do this here:

```suggestion
                if let authContext = coordinator?.authContext {
                    let profileViewModel = RemoteProfileViewModel(
                        context: AppContext.shared,
                        authContext: authContext,
                        acct: components[1]
                    )
                    self.coordinator?.present(
                        scene: .profile(viewModel: profileViewModel),
                        from: nil,
                        transition: .show
                    )
                }
 ```

File: Mastodon/Scene/Profile/RemoteProfileViewModel.swift [hunk 9]
```diff
@@ -46,7 +46,9 @@ final class RemoteProfileViewModel: ProfileViewModel {
                     assertionFailure()
                     return
                 }
-                self.user = mastodonUser
+                DispatchQueue.main.async {
```
DK1DA: As we're currently expecting the `receiveValue` to be running on the main thread anyways, we could also add a `            .receive(on: DispatchQueue.main)` to the combine flow.

kimar (changes requested):
Thank you very much @jinsu35 ! I'd really like to get this in as I'm working on a feature that would build upon your work. However I think your implementation can be done without the changes to`APIService` which kind break our current api conventions (e.g. returning `Mastodon.Response.Content<T>`).

Fetching status only works for local statuses right now as far as I can see, but we can improve this later for sure.

Let me know if there are any questions or if I can help.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Account.swift
jinsu35: We can't use `accountLookup()` as it is now because it does not create a Persistent record. In case a record doesn't exist, we need to create one upon lookup. So I'm adding a new function `accountSearch()` that does the lookup and creates a record.

DK1DA: Ah fair enough, that makes sense. I guess for me it worked because I was using your `accountSearch` method which already persisted the object for me.

kimar (approved):
Thank you very much @jinsu35 this helps a lot with what I'm currently building.

frogg:
Amazing, thanks a lot for building this! @jinsu35

timonus:
Does "status" only work for status IDs in your instance?

asci:
Hey folks, do I understand correctly that it is not possible to show status from a different instance than a home one due to ID mismatch and it is not possible to add some initial text to post (like it is possible with `/share?url=text`. Thanks!

PR #853, opened 2023-01-03
Create a workflow to auto-update generated files

j-f1:
This workflow runs on the automatically generated (?) PRs from Crowdin like #804, runs `./update_localization.sh`, and pushes the result. This avoids a situation where the generated files are out of sync with the latest localization strings. Since it only runs on the `l10n_develop` branch, any code changes it makes can be reviewed as part of the normal code review process — it won’t make changes to `main`.


j-f1:
Not sure if there’s a way to fix the above issue. Willing to attribute the commit to a human but ideally bots could be exempted from the CLA

PR #856, opened 2023-01-04
feat(profile): Implement tap to copy username

kimar:
# Rationale

Implement a way to copy the username from a profile by showing a menu after tapping the username.

# Demo

RocketSim_Recording_iPhone_14_Pro_2023-01-04_16 37 34

### DIFF
--- a/Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
+++ b/Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
@@ -172,15 +172,26 @@ final class ProfileHeaderView: UIView {
         textField.autocapitalizationType = .none
         return textField
     }()
-
-    let usernameLabel: UILabel = {
-        let label = UILabel()
-        label.font = UIFontMetrics(forTextStyle: .callout).scaledFont(for: .systemFont(ofSize: 16, weight: .regular))
-        label.adjustsFontSizeToFitWidth = true
-        label.minimumScaleFactor = 0.5
-        label.textColor = Asset.Colors.Label.secondary.color
-        label.text = "@alice"
-        return label
+    
+    private lazy var usernameButtonMenu: UIMenu = {
+        UIMenu(children: [
+            UIAction(title: L10n.Common.Controls.Actions.copy, handler: { [weak self] _ in
+                UIPasteboard.general.string = self?.usernameButton.title(for: .normal)
+            })
+        ])
+    }()
+    
+    lazy var usernameButton: UIButton = {
+        let button = UIButton()
+        button.setTitle("@alice", for: .normal)
+        button.titleLabel?.font = UIFontMetrics(forTextStyle: .callout).scaledFont(for: .systemFont(ofSize: 16, weight: .regular))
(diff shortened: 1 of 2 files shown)


Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
```diff
@@ -172,15 +172,26 @@ final class ProfileHeaderView: UIView {
...
-        label.font = UIFontMetrics(forTextStyle: .callout).scaledFont(for: .systemFont(ofSize: 16, weight: .regular))
-        label.adjustsFontSizeToFitWidth = true
-        label.minimumScaleFactor = 0.5
-        label.textColor = Asset.Colors.Label.secondary.color
-        label.text = "@alice"
-        return label
+    
+    private lazy var usernameButtonMenu: UIMenu = {
+        UIMenu(children: [
+            UIAction(title: L10n.Common.Controls.Actions.copy, handler: { [weak self] _ in
```
j-f1: Two thoughts:

1. Should this use the [new API for edit menus](https://developer.apple.com/documentation/uikit/uieditmenuinteraction/) in iOS 16? (so that users get the little mini popover thing like when you select text)
2. Can you add a `doc.on.doc` icon so the menu doesn’t have the text alone?

zeitschlag: I disagree with 1., as we still have to support iOS 15 and another `@available` with two different implementations is not worth the effort in my opinion. But I like 2.!

Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
```diff
@@ -417,7 +428,11 @@ extension ProfileHeaderView {
         // nameMetaText.textView.setContentHuggingPriority(, for: <#T##NSLayoutConstraint.Axis#>)
         
         nameContainerStackView.addArrangedSubview(displayNameStackView)
-        nameContainerStackView.addArrangedSubview(usernameLabel)
+        nameContainerStackView.addArrangedSubview(usernameButton)
+        
+        NSLayoutConstraint.activate([
+            usernameButton.heightAnchor.constraint(equalToConstant: 20)
```
zeitschlag: Out of curiousity: Have you checked what happens, when users use Dynamic Type? Is the text still visible although the height-constraint is limited to `20`?

zeitschlag (approved):

Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
DK1DA: Very fair point! Should use `greaterThanOrEqualToConstant` probably. But anyways there is currently no change in behavior, as `adjustsFontSizeToFitWidth = true` leads to the username always being shrinked to fit the label width, even in the current release.

Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
DK1DA: Very good point @j-f1 , added `doc.on.doc`.

PR #866, opened 2023-01-07
Fix correcting image rotation

woxtu:
Hi,

The `normalized` method that corrects image rotation returns a wrong size image, so this PR fixes that.


j-f1:
I think the current behavior is correct — I’ve encountered this before when taking pictures with the camera rotated. The native resolution of the camera would be 640×480, for example, and I would take a portrait picture. The resulting image would still be 640×480, but the orientation property would be changed, and the image should always be rendered rotated as 480×640. Is my interpretation missing something?

woxtu:
Thanks, I misunderstood the function of the method.

PR #873, opened 2023-01-09
fix(share-menu): Fixes URL for shared status when using square.and.arrow.up share menu on status

kimar:
# Expected Behavior

Sharing a link always uses the permalink of the post.

# Actual Behavior

Sharing a link from a reblog uses the permalink of the reblog (which kind of doesn't exist).

# Steps to Reproduce the Problem
(Using the share menu on the upper right hand side of a status)

* Have a post in timeline that is a reblog
* Click share

# The Fix

Currently the `menuContext.status` is being used, which is the actual reblog status and not the reblogged item's status. This is valid in most cases, such as muting the reblogging user etc. but it's not useful when sharing the status as you want to share the shared status.


zeitschlag (approved):
LGTM

PR #877, opened 2023-01-10
Left-align the username in the profile header

j-f1:
The changes to add a menu to the username so it can be copied also center-aligned it under the display name. This PR overrides the text alignment so that it returns to its previous position.


kimar (approved):
Great catch @j-f1 ! Many thanks for the fix! 👏

PR #878, opened 2023-01-11
Feature: Action Extension

kimar:
# Rationale

Implement an Action Extension that can be used to e.g. directly open a user profile in the app from the browser.

# Details

If you are on a site that's detected as a profile then the Profile will be opened in the App on the Profile details screen.

If the link is another link that is a valid mastodon link then it'll always open the search with the link as the search content, this usually leads to the expected results.

# Demo

RocketSim_Recording_iPhone_14_Pro_2023-01-11_14 52 28


OpenInActionExtension/Action.js
```diff
@@ -0,0 +1,47 @@
...
+            alert(alertMessage)
+        } else {
+            window.location = arguments["openURL"]
+        }
+    }
+    
+};
+
+function detectUsername() {
+    var uriUsername = document.documentURI.match("@(.+)@([a-z0-9]+\.[a-z0-9]+)")
```
zeitschlag: More of a question: Is this like the _real_ regex to get Mastodon-usernames? I'm asking, as something like `@mailto:email😊@mastodon.social@mastodon.social` (or worse, like crashing strings) would be valid, too, I guess?

I'm talking about the `(.+)`. But to be honest, I have no idea, which characters a valid, I was expecting something like `[a-zA-Z0-9.]` here.

[Special characters in usernames](https://github.com/mastodon/mastodon/issues/6830) seem to be a hot topic

OpenInActionExtension/Media.xcassets/Icon.appiconset/Contents.json
```diff
@@ -0,0 +1,14 @@
+{
+  "images" : [
+    {
+      "filename" : "MastodonActionExtensionIcon@3x.png",
```
zeitschlag, line 4: What's the reason for `.png` instead of `.pdf` or another vector-graphic-format?

OpenInActionExtension/Media.xcassets/TouchBarBezel.colorset/Contents.json
```diff
@@ -0,0 +1,14 @@
+{
+  "info" : {
+    "version" : 1,
+    "author" : "xcode"
+  },
+  "colors" : [
+    {
+      "idiom" : "mac",
+      "color" : {
+        "reference" : "systemPurpleColor"
```
zeitschlag: We could use our very own blurple here, if we can (which I don't know)?

zeitschlag (approved):
Nice!

OpenInActionExtension/Media.xcassets/Icon.appiconset/Contents.json
zeitschlag: Ah, nevermind, just found the archive

OpenInActionExtension/Media.xcassets/TouchBarBezel.colorset/Contents.json
DK1DA: Yes, makes total sense 👍  I completely ignore this color. Fixed!

OpenInActionExtension/Action.js
DK1DA: I've picked this deliberately because I wasn't sure if a username can't contain anything else than ascii chars, but that said, maybe it's a good idea to double check and this is what we're using in the iOS app:

`"(?:@([a-zA-Z0-9_]+)(@[a-zA-Z0-9_.-]+)?|#([^\\s.]+))"`

So I'll repurpose the same in the JS code as well.

PR #879, opened 2023-01-12
Add template for header

zeitschlag:
This changes the header at the beginning of a file when you create a new file.

In other words: The PR changes this:

grafik

into this:

grafik

All the other information (When was this file created? And by whom?) is stored in git. This is more or less a proposal, lmk, when something should be different.


kimar (approved):
I really like the idea. 👏

PR #881, opened 2023-01-12
Apply languageIdentifier attribute to user-generated content where possible

j-f1:
[builds on #880, feel free to merge just this one to get the changes from both]

This allows VoiceOver to dynamically switch voices based on the language a post was written in, without affecting the pronunciation of UI elements. For example, if reading a German post while the device is set to English, the English voice would read the username, date, “# media,” and “Actions available”, while the German voice would read the actual post content. This also applies to alt text (which inherits the language of the post).

It would be nice to also read people’s names and bios in the appropriate language, but that’s blocked on https://github.com/mastodon/mastodon/issues/23067 which I’ve just opened.

### DIFF
--- a/MastodonSDK/Sources/CoreDataStack/Entity/Mastodon/StatusEdit.swift
+++ b/MastodonSDK/Sources/CoreDataStack/Entity/Mastodon/StatusEdit.swift
@@ -31,6 +31,9 @@ public final class StatusEdit: NSManagedObject {
     // sourcery: autoUpdatableObject, autoGenerateProperty
     @NSManaged public var spoilerText: String?
 
+    // sourcery: autoUpdatableObject, autoGenerateProperty
+    @NSManaged public var status: Status?
+
     // MARK: - AutoGenerateProperty
     // sourcery:inline:StatusEdit.AutoGenerateProperty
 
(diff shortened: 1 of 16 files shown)


j-f1:
~This needs work — when the post is translated it should switch to the current UI language.~

MastodonSDK/Sources/CoreDataStack/Entity/Mastodon/StatusEdit.swift
```diff
@@ -31,6 +31,9 @@ public final class StatusEdit: NSManagedObject {
     // sourcery: autoUpdatableObject, autoGenerateProperty
     @NSManaged public var spoilerText: String?
 
+    // sourcery: autoUpdatableObject, autoGenerateProperty
+    @NSManaged public var status: Status?
```
DK1DA, line 35: ❓ Does the CoreData model already reflect this relationship? I haven't seen any changes to our Model in this PR.

j-f1: Yes:

https://github.com/mastodon/mastodon-ios/blob/d24fe00420849f008a3301a754f7d601fdcf63e3/MastodonSDK/Sources/CoreDataStack/CoreData.xcdatamodeld/CoreData%208.xcdatamodel/contents#L255

PR #884, opened 2023-01-13
Fix broken i18n files

j-f1:
This fixes some issues with the translation files that were causing translations to be deleted upon a rebuild. In the future, only `Localization/app.json` should be manually edited, and then all other files will be generated from that source.


j-f1:
Looks like this is sorted now!

PR #888, opened 2023-01-16
Improve "Open in Mastodon" by using Search API

kimar:
# Rationale

Previously the search action extension was failing on some occasion due to the JS generated DOM, of the links in question which the user might wanted to open. We're now leveraging the mastodon Search API to make sense of the web URLs.

### DIFF
--- a/OpenInActionExtension/ActionRequestHandler.swift
+++ b/OpenInActionExtension/ActionRequestHandler.swift
@@ -83,7 +83,7 @@ private extension ActionRequestHandler {
             )
             .receive(on: DispatchQueue.main)
             .sink { completion in
-                
+                // no-op
             } receiveValue: { [weak self] result in
                 let value = result.value
                 if let foundAccount = value.accounts.first {
--- a/OpenInActionExtension/ActionRequestHandler.swift
+++ b/OpenInActionExtension/ActionRequestHandler.swift
@@ -9,6 +9,7 @@ import Combine
 import UIKit
 import MobileCoreServices
 import UniformTypeIdentifiers
+import MastodonCore
 import MastodonSDK
 import MastodonLocalization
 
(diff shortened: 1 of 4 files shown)


OpenInActionExtension/ActionRequestHandler.swift
```diff
@@ -56,13 +60,53 @@ class ActionRequestHandler: NSObject, NSExtensionRequestHandling {
...
+            let activeAuthenticationBox = Self.appContext
+                .authenticationService
+                .mastodonAuthenticationBoxes
+                .first
+        else {
+            return doneWithResults(nil)
+        }
+        
+        Mastodon.API
+            .V2
```
j-f1, line 76: Is there a way to handle folks running old versions of Mastodon without the v2 search API?

zeitschlag: If someone still runs [Mastodon with a version lower than 3.0](https://github.com/mastodon/mastodon/releases/tag/v3.0.0), they have way more severe issues than a broken app :D

OpenInActionExtension/ActionRequestHandler.swift
```diff
@@ -9,13 +9,19 @@ import Combine
...
 import MastodonSDK
 import MastodonLocalization
 
 class ActionRequestHandler: NSObject, NSExtensionRequestHandling {
     var extensionContext: NSExtensionContext?
     var cancellables = [AnyCancellable]()
     
+    /// Capturing a static shared instance of AppContext here as otherwise there
+    /// will be lifecycle issues and we don't want to keep multiple AppContexts around
+    /// in case there another Action Extension process is spawned
```
zeitschlag, line 22: 👍 for this comment!

zeitschlag (approved):
LGTM

PR #891, opened 2023-01-17
Added a method to clear the cache and to check if a key exists in cache.

CHIRAGGARORA:


Gargron:
What for? This is not used anywhere?

PR #900, opened 2023-01-20
Intent: Set the `url` property of the Post object before returning it

### LINKED ISSUE #899: [BUG] Shortcut "Post on Mastodon" action returned `Post` result's URL doesn't work

### Current Behavior

IMG_6F9B061FD3F5-1

This is an example Shortcut workflow. The posting works, but the return result, the `Post` doesn't seem to be able to get the URL of the just created post.

### Expected Behavior

I might be guessing wrongly. My understanding is that the `URL` should return the new post's mastodon URL.

### Steps To Reproduce

1. create a Shortcut workflow like I did above
2. run the workflow
3. filling some test text
4. post it to mastodon
5. the last step will only have the text, without URL.

### Environment

```markdown
- Device: iPhone 13 Pro
- OS: 16.2
- Version: 1.4.9
- Build: 224
```

### Anything else?

_No response_

j-f1:
Fixes #899.

### DIFF
--- a/MastodonIntent/Handler/SendPostIntentHandler.swift
+++ b/MastodonIntent/Handler/SendPostIntentHandler.swift
@@ -97,8 +97,8 @@ extension SendPostIntentHandler: SendPostIntentHandling {
                     subtitle: content,
                     image: response.value.account.avatarImageURL().flatMap { INImage(url: $0) }
                 )
-                if let urlString = response.value.url {
-                    post.url = URL(string: urlString)
+                if let urlString = response.value.url, let url = URL(string: urlString) {
+                    post.url = url
                 }
                 posts.append(post)
             }   // end for in
--- a/MastodonIntent/Handler/SendPostIntentHandler.swift
+++ b/MastodonIntent/Handler/SendPostIntentHandler.swift
@@ -97,6 +97,7 @@ extension SendPostIntentHandler: SendPostIntentHandling {
                     subtitle: content,
                     image: response.value.account.avatarImageURL().flatMap { INImage(url: $0) }
                 )
+                post.url = URL(string: response.value.url ?? "")
                 posts.append(post)
             }   // end for in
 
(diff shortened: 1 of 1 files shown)


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,15 @@
...
+    "@glimmer/syntax",
+    "bear-check-shared",
+    "glimmer-engine",
+    "prettier"
+  ],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
zeitschlag: What is this? 🙈

MastodonIntent/Handler/SendPostIntentHandler.swift
```diff
@@ -97,6 +97,7 @@ extension SendPostIntentHandler: SendPostIntentHandling {
                     subtitle: content,
                     image: response.value.account.avatarImageURL().flatMap { INImage(url: $0) }
                 )
+                post.url = URL(string: response.value.url ?? "")
```
zeitschlag: Maybe it makes sense to use an `if let` here, if `response.value.url` is invalid and not add the URL instead of adding a URL with an empty string? I don't have any hard feelings on this, just an idea :)

node_modules/.yarn-integrity
j-f1: Oops, removed! Not sure where that came from.

zeitschlag (approved):

MastodonIntent/Handler/SendPostIntentHandler.swift
j-f1: That’s definitely better, I used the fact that `URL(string: "")` returns nil here but that’s not very understandable (and it definitely took me a second to remember that’s what happens when looking at the code again). Going with your approach 👍

zeitschlag: Oh, you're absolutely right, I didn't know that, so thank you :)

MastodonIntent/Handler/SendPostIntentHandler.swift
```diff
@@ -97,6 +97,9 @@ extension SendPostIntentHandler: SendPostIntentHandling {
                     subtitle: content,
                     image: response.value.account.avatarImageURL().flatMap { INImage(url: $0) }
                 )
+                if let urlString = response.value.url {
```
zeitschlag: I meant something like this:

```swift
if let urlString = response.value.url, let url = URL(string: urlString) {
    post.url = url
}
```

🙈

PR #907, opened 2023-01-23
Add support for selecting the post language

### LINKED ISSUE #508: [BUG] Missing language selector

### Current Behavior

Language selector is not available in the “New Post” screen. 

This is a must-have for people who toot in multiple languages and wish to provide the correct language metadata to filters and screen readers. 

### Expected Behavior

When composing a new post, users should be able to select a language other than the default one in their preferences. 

### Steps To Reproduce

_No response_

### Environment

```markdown
- Device: iPhone 14 Pro
- OS: iOS 16.1
- Version: 1.4.5
- Build: 144
```

### Anything else?

_No response_

j-f1:
Fixes #508. A new button is added to the toolbar. It opens a menu when tapped with up to 4 sections:

- “Recent:” up to 3 languages you’ve previously made a post in using the app
- “Suggested:” a local system-provided natural language model is run on the post text in the background, and any detected languages are displayed here
- if the currently selected language isn’t in either of the above sections, it will be displayed separately.
- A “Other Language…” button opens a sheet/popover with a full list of languages (currently the list is provided by the OS since there is no API route to fetch the full set of valid languages afaict). Searching works across both the language code and the text displayed in the list. Tapping on a language selects it.

There’s not really a good way to show 3 letter language codes, especially since the letters can be in a variety of widths:

Additionally, a badge is added to the language picker if we have high confidence (> 0.99) that the currently selected language does not match the language of the post content. Selecting any language, even the currently selected one, dismisses the badge until you start a new post.

Related: #704 (would need to add language support to that PR after merging this one or vice versa)


samhenrigold:
(Mastodon product designer here) 
Thanks for making this PR! Got one small suggestion if you don't mind...

Regarding three-letter lang codes, I wonder if we could use a [compressed font style](https://developer.apple.com/documentation/uikit/uifont/width)? I know it's only ≤ 16.x but it could give us some wiggle room:
Screenshot 2023-01-23 at 1 51 13 PM
(In this mockup, I'm using 11pt semibold for 2 character, 10pt semibold compressed for 3 characters)

j-f1:
That’s a great idea! Doing the compressed font style greatly helps with the 3 letter case. Trying to wrangle SwiftUI to get shrink-to-fit to work since the variation in letter widths is long enough that a single font size per character count alone won’t work to nicely size all language codes.

j-f1:
Great idea! The compressed width is definitely the solution. I’m not sure if going up to 11pt will work for the 2-character names since `AM` is pretty tight already at 10pt and collides with the borders at 11. However, the 3 letter codes look awesome! 



Slapping a `minimumScaleFactor` on the text (in addition to setting the font width property on iOS 16) fixed the issues with wider letters: (before/after)

.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
...
+            "type": "lldb",
+            "request": "launch",
+            "name": "Release StringsConvertor (Localization/StringsConvertor)",
+            "program": "${workspaceFolder:mastodon-ios}/Localization/StringsConvertor/.build/release/StringsConvertor",
+            "args": [],
+            "cwd": "${workspaceFolder:mastodon-ios}/Localization/StringsConvertor",
+            "preLaunchTask": "swift: Build Release StringsConvertor (Localization/StringsConvertor)"
+        }
+    ]
+}
```
zeitschlag: What does a vscode config have to do with Post-language? And who takes care of this config in the future? :P

j-f1: oops, that’s an accidental add.

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/ComposeContentToolbarView+ViewModel.swift
```diff
@@ -83,6 +91,8 @@ extension ComposeContentToolbarView.ViewModel {
                 return Asset.Scene.Compose.chatWarning.image.withRenderingMode(.alwaysTemplate)
             case .visibility:
                 return Asset.Scene.Compose.earth.image.withRenderingMode(.alwaysTemplate)
+            case .language:
+                fatalError()
```
zeitschlag: Tbh: I have a bad gut feeling with these `fatalError()`, but I don't have an idea, how to do it better in this case. From my perspective we must have a `UIImage` and can't use a `UIImage?` here, right?

j-f1: Yeah. An alternative that I’d be happy to do here or in a separate PR would be to remove the computed variables here and instead define the appropriate properties in the view…

```swift
case .attachment:
  Menu {
    // ...
  } label: {
    ComposeContentToolbarLabel(image: Asset.Scene.Compose.media)
  }.disabled(!viewModel.isAttachmentButtonEnabled)
case .poll:
  Button {
    // ...
  } label: {
    ComposeContentToolbarLabel(image: viewModel.isPollActive ? Asset.Scene.Compose.pollFill : Asset.Scene.Compose.poll)
  }.disabled(!viewModel.isPollButtonEnabled)
```

zeitschlag: Maybe for now you can just add a short message to the `fatalError()`, why this should never ever happen? For the bigger refactoring I'd suggest to use its own PR. What do you think?

j-f1: Sure!

j-f1: Done. Will open the other PR shortly!

zeitschlag (approved):

PR #911, opened 2023-01-24
Deprecation note for UIButton.setInsets(forContentPadding:imageTitlePadding:)

rmnli:
### Problem

`UIButton.setInsets(forContentPadding:imageTitlePadding:)` hides the warnings about UIButton.contentEdgeInsets and UIButton.titleEdgeInsets. 

### Impact of the change

All uses of the function will be highlighted by Xcode. The developer will be informed that the function includes the deprecated API.

### DIFF
--- a/MastodonSDK/Sources/MastodonExtension/UIButton.swift
+++ b/MastodonSDK/Sources/MastodonExtension/UIButton.swift
@@ -8,6 +8,7 @@
 import UIKit
 
 extension UIButton {
+    @available(iOS, deprecated: 15.0, message: "This method is ignored when using UIButtonConfiguration. The effect of this method can be replicated via UIButtonConfiguration.contentInset and UIButtonConfiguration.imagePadding.")
     public func setInsets(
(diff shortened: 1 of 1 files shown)


j-f1:
Per https://github.com/mastodon/mastodon-ios/pull/783#issuecomment-1356461118, these warnings are probably safe to ignore since none of the code is using the `UIButton.Configuration` API.

rmnli:
@j-f1 thanks for clarification.
But I noticed that the new api is already in use (see [Scene/Onboarding/ConfirmEmail/MastodonConfirmEmailViewController.swift](https://github.com/mastodon/mastodon-ios/blob/0a9689c67f3bc471a0e927512d6e16cdc9da483a/Mastodon/Scene/Onboarding/ConfirmEmail/MastodonConfirmEmailViewController.swift)).
And I agree that this is safe, but it will be harder to migrate in the future if `UIButton.setInsets` is used frequently in the project.

j-f1:
Good point, I guess it makes sense to leave the warnings in then!

PR #914, opened 2023-02-03
chore(Feed): Improve Feed performance by fetching statuses in batches

kimar:


j-f1:
Nice! I think this will also make the scroll bar more useful since right now when I grab it and move, like, one pixel, I scroll a whole week back.

j-f1:
- Also, related: https://github.com/mastodon/mastodon-ios/pull/817

kimar:
@j-f1 thank you! I'll also take a look at #817 but for now I'll try to see if this tiny PR here can already improve the initial loading UX. Currently the app easily takes 2s until it renders some content, and I don't see a reason why we should lots thousands (in my case 24k) statuses at once.

MastodonSDK/Sources/MastodonCore/FetchedResultsController/FeedFetchedResultsController.swift
```diff
@@ -16,18 +16,31 @@ import MastodonSDK
...
     // input
     @Published public var predicate = Feed.predicate(kind: .none, acct: .none)
     
     // output
     private let _objectIDs = PassthroughSubject<[NSManagedObjectID], Never>()
     @Published public var records: [ManagedObjectRecord<Feed>] = []
+
+    public func fetchNextBatch() {
+        fetchedResultsController.fetchRequest.fetchLimit += Constants.defaultFetchLimit
+        try? fetchedResultsController.performFetch()
```
j-f1, line 42: A couple of questions:

- how does this handle insertion of new posts at the top of the timeline (or loading missing posts at a gap)?
- is it bad to re-fetch the full list of posts even if you’ve reached the very bottom of your timeline?

DK1DA: Regarding
1. this should not have an impact, once the diffable datasource determines there is a gap between statuses the option to load the missing will be shown. The behavior of the diffable datasource should not be changed right now
2. That's a good question, DB wise we're fetching the items of course, from a UX perspective there shouldn't be any UI rerender happening as the diffable datasource won't detect any changes.

zeitschlag (approved):

PR #916, opened 2023-02-07
IOS-67: Add the “Header” and “Button” traits to the logo button

j-f1:
I added the “header” and “button” traits but could not find a “label” trait. There is “[static text](https://developer.apple.com/documentation/uikit/uiaccessibilitytraits/1620206-statictext)” but that doesn’t feel appropriate for a button.


kimar (approved):
@j-f1 first of all thank you again for your contributions! I think they weren't aware that this is actually a button and not a label, so I think the traits you picked are perfectly fine.

PR #917, opened 2023-02-07
IOS-81: Make “Translated From” UI accessible

j-f1:
Fixes IOS-81.

- Added “Show Original” action in threads/feeds
- Added “Translated from X using Y” to the accessibility label in threads/feeds
- Added the “Translated from X using Y / Show Original” button as an accessibility action on the main post in a thread


kimar (approved):
👏  LGTM!

PR #918, opened 2023-02-07
IOS-77: Improve accessibility of account rows in search results/recent searches

j-f1:
Also fixes a layout issue that could cause clipping of very long display names / handles


kimar (approved):
Great! Thanks a bunch! ❤️

PR #920, opened 2023-02-07
IOS-73: Mark AvatarButton as an image

j-f1:
VoiceOver now describes the avatar! “a person smiling and posing for a photo in front of a multicolored background” You can even swipe to access the “Explore Image” action and this UI shows up: (example: `@QasimRashid@mastodon.social`)


kimar (approved):
Ok wow 🤯  this is actually terrific what you can achieve by adding a single line of code.

PR #922, opened 2023-02-07
feat(App): Implement Haptic Feedback for Tab switches and Timeline reload

kimar:
# Rationale

Implements haptic feedback for the following scenarios:

- [x] Main tabs
- [x] Pull to refresh (implemented by UIRefreshControl natively)
- [x] When new content loads without user action


zeitschlag (approved):

j-f1:
The timeline reload seems to happen in the background and the haptic feedback feels very similar to the tap I get when favoriting or boosting a post, which has made me worry that I accidentally did one of those actions.

PR #923, opened 2023-02-07
IOS-76: Improve accessibility of the About fields on profiles

j-f1:
Actioning the item will open the first link, and you can swipe through all of the links as secondary actions.


kimar (approved):
👍  🚀

PR #924, opened 2023-02-07
IOS-72: Improve accessibility for the poll composer UI

j-f1:
- Add an appropriate label to the “Add Option” button
- Add secondary actions for “move up” / “move down” / “remove option” where appropriate
- Hide the circle and grabber icons from accessibility tools
- Mark the whole poll view as a group so that when VoiceOver enters it it reads “Poll, Option 1” for example


MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Poll/PollOptionRow.swift
```diff
@@ -38,12 +43,67 @@ public struct PollOptionRow: View {
...
+                        }
+                        if let moveDown {
+                            Button(L10n.Scene.Compose.Poll.moveDown, action: moveDown)
+                        }
+                        if let removeOption {
+                            Button(L10n.Scene.Compose.Poll.removeOption, action: removeOption)
+                        }
+                    }
+                } else {
+                    // beautiful!
```
DK1DA: @j-f1 I allowed myself to refactor this, I hope that's ok for you, it's not meant offensive I just also didn't want to consume your time too much as you're already doing a fabulous job.

j-f1: That would be great! I couldn’t think of a better way to do this, would love to see what you came up with

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Poll/PollOptionRow.swift
```diff
@@ -38,12 +43,55 @@ public struct PollOptionRow: View {
...
+                        }
+                        if let moveDown {
+                            Button(L10n.Scene.Compose.Poll.moveDown, action: moveDown)
+                        }
+                        if let removeOption {
+                            Button(L10n.Scene.Compose.Poll.removeOption, action: removeOption)
+                        }
+                    }
+                } else {
+                    switch (moveUp, moveDown, removeOption) {
```
DK1DA, line 60: @j-f1 I've used switch pattern matching here and I think it's fine. Thanks for your help!

kimar (approved):
LGTM 👍

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Poll/PollOptionRow.swift
j-f1: That looks good! And when we drop iOS 15 support the whole switch can be removed.

DK1DA: The good thing is also that you basically don't forget to cover cases as the compiler would otherwise complain about the switch statement.

PR #927, opened 2023-02-08
IOS-92: Enlarge post action buttons at larger text sizes

j-f1:
The share icon does not appear aligned with the boost button because the images are slightly different widths. ¯\\\_(ツ)\_/¯

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
@@ -64,6 +64,7 @@ extension ActionToolbarContainer {
         reblogButton.addTarget(self, action: #selector(ActionToolbarContainer.buttonDidPressed(_:)), for: .touchUpInside)
         favoriteButton.addTarget(self, action: #selector(ActionToolbarContainer.buttonDidPressed(_:)), for: .touchUpInside)
         shareButton.addTarget(self, action: #selector(ActionToolbarContainer.buttonDidPressed(_:)), for: .touchUpInside)
+        addInteraction(UILargeContentViewerInteraction())
 
         container.arrangedSubviews.forEach { subview in
             container.removeArrangedSubview(subview)
@@ -78,6 +79,7 @@ extension ActionToolbarContainer {
             button.setTitleColor(.secondaryLabel, for: .normal)
             button.expandEdgeInsets = UIEdgeInsets(top: -10, left: -10, bottom: -10, right: -10)
             button.setInsets(forContentPadding: .zero, imageTitlePadding: 4)
+            button.showsLargeContentViewer = true
         }
         // add more expand for menu button
         shareButton.expandEdgeInsets = UIEdgeInsets(top: -10, left: -20, bottom: -10, right: -20)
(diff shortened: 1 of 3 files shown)


MastodonSDK/Package.resolved
```diff
@@ -72,15 +72,6 @@
...
     },
-    {
-      "identity" : "kingfisher",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/onevcat/Kingfisher.git",
-      "state" : {
-        "revision" : "44e891bdb61426a95e31492a67c7c0dfad1f87c5",
-        "version" : "7.4.1"
-      }
-    },
```
zeitschlag: These two changes (removing `kingfisher` and `thirdpartymailer` from `Package.resolved`) is just cleanup, right?

MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
```diff
@@ -158,22 +127,9 @@ extension ActionToolbarContainer {
         case reply
         case reblog
         case like
-        case bookmark
```
zeitschlag, line 161: What's the need to remove this?

MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
```diff
@@ -85,7 +78,8 @@ extension ActionToolbarContainer {
             button.setTitle("", for: .normal)
             button.setTitleColor(.secondaryLabel, for: .normal)
             button.expandEdgeInsets = UIEdgeInsets(top: -10, left: -10, bottom: -10, right: -10)
-            button.setInsets(forContentPadding: .zero, imageTitlePadding: style.buttonTitleImagePadding)
+            button.setInsets(forContentPadding: .zero, imageTitlePadding: 4)
+            button.showsLargeContentViewer = true
```
zeitschlag: The `showsLargeContentViewer` seems to be a workaround, right?

What I would have expected from this issue/PR is Dynamic Type-support for the four buttons (reply, reblog, favorite, share). Not that `UILargeContentViewerInteraction` is a bad thing, but it's only available for the huge accessibility Dynamic Type-values.

The font itself doesn't change: 

```swift
button.titleLabel?.font = .monospacedDigitSystemFont(ofSize: 12, weight: .regular)
```

MastodonSDK/Package.resolved
j-f1: Yes, I believe so — Xcode did this automatically.

MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
j-f1: Correct, as I said in the OP the images cannot be easily resized in UIKit, but I could just resize the text.

MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
j-f1: The button no longer exists so I removed the dead code.

MastodonSDK/Sources/MastodonUI/View/Control/ActionToolbarContainer.swift
j-f1: Update: figured out how to get the images to resize, but having some fun getting the icons to break onto 2 lines for larger text sizes while still retaining the current look on smaller text sizes

j-f1:
Got it sorted partially thanks to https://mastodon.social/@dadederk@iosdev.space/110044528429919745!

kimar (approved):
LGTM 💯

PR #929, opened 2023-02-08
Change random server picking logic to require "general" category, prefer no approval and own language code

kimar:
# Rationale

The following logic is desired:

```
general purpose (must), instant sign-up (should), matching language (should)
```

### DIFF
--- a/Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
+++ b/Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
@@ -247,7 +247,7 @@ extension MastodonPickServerViewModel {
         let languageMatchesWithoutApproval = noApprovalRequired.filter { $0.language.lowercased() == language }
         let languageMatchesWithApproval = approvalRequired.filter { $0.language.lowercased() == language }
         let languageDoesNotMatchWithoutApproval = noApprovalRequired.filter { $0.language.lowercased() != language }
-        let languageDoesNotMatchWithApproval = noApprovalRequired.filter { $0.language.lowercased() != language }
+        let languageDoesNotMatchWithApproval = approvalRequired.filter { $0.language.lowercased() != language }
 
         switch (
             languageMatchesWithoutApproval.isEmpty,
--- a/Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
+++ b/Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
@@ -235,9 +235,37 @@ extension MastodonPickServerViewModel {
         let servers = indexedServers.value
         guard servers.isNotEmpty else { return nil }
 
-        let randomServer = servers.filter {
-            $0.language.lowercased() == language
-        }.randomElement()
+        let generalServers = servers.filter {
+            $0.categories.contains("general")
+        }
+        
+        let randomServer: Mastodon.Entity.Server?
+        
+        let noApprovalRequired = generalServers.filter { !$0.approvalRequired }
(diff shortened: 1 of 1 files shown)


Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
```diff
@@ -235,9 +235,37 @@ extension MastodonPickServerViewModel {
...
+        
+        let randomServer: Mastodon.Entity.Server?
+        
+        let noApprovalRequired = generalServers.filter { !$0.approvalRequired }
+        let approvalRequired = generalServers.filter { $0.approvalRequired }
+        
+        let languageMatchesWithoutApproval = noApprovalRequired.filter { $0.language.lowercased() == language }
+        let languageMatchesWithApproval = approvalRequired.filter { $0.language.lowercased() == language }
+        let languageDoesNotMatchWithoutApproval = noApprovalRequired.filter { $0.language.lowercased() != language }
+        let languageDoesNotMatchWithApproval = noApprovalRequired.filter { $0.language.lowercased() != language }
```
zeitschlag: Those are the same:

```swift
noApprovalRequired.filter { $0.language.lowercased() != language }
noApprovalRequired.filter { $0.language.lowercased() != language }
```

From my understanding (and variable-names) I'd have suspected something like this:

```swift
noApprovalRequired.filter { $0.language.lowercased() != language }
approvalRequired.filter { $0.language.lowercased() != language }
```

Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
```diff
@@ -235,9 +235,37 @@ extension MastodonPickServerViewModel {
...
+
+        switch (
+            languageMatchesWithoutApproval.isEmpty,
+            languageMatchesWithApproval.isEmpty,
+            languageDoesNotMatchWithoutApproval.isEmpty,
+            languageDoesNotMatchWithApproval.isEmpty
+        ) {
```
zeitschlag, line 257: I already said it, but: I like this tuple/pattern-matching-approach 👍

zeitschlag (changes requested):
See first comment :)

Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
```diff
@@ -235,9 +235,37 @@ extension MastodonPickServerViewModel {
...
+            randomServer = languageDoesNotMatchWithApproval.randomElement()
+        case (true, true, false, _):
+            randomServer = languageDoesNotMatchWithoutApproval.randomElement()
+        case (true, false, _, _):
+            randomServer = languageMatchesWithApproval.randomElement()
+        case (false, _, _, _):
+            randomServer = languageMatchesWithoutApproval.randomElement()
+        }
 
         return randomServer ?? servers.randomElement() ?? servers.first
```
zeitschlag, line 270: If everything goes wrong, this fallback could still lead people to a small instance, but from my perspective, this is not very likely.

Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
DK1DA: Yep you're right, thanks you! fixed.

Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
DK1DA: Yes I think that's expected in the worst case

zeitschlag (approved):
Thank you!

PR #933, opened 2023-02-09
Improve the UX around post translation

j-f1:
- move the activity indicator above the link card (since the card is not translated)
- stop showing the activity indicator when selecting actions other than “Translate from X”
- show the activity indicator on the root cell of threads
- improve some layout issues


kimar (approved):
👏  great, thanks @j-f1

PR #875, opened 2023-01-09
Implement post editing / edit history

zeitschlag:
# Rationale

| Demo | Description |
|---|---|
| RocketSim_Recording_iPhone_14_Pro_2023-02-23_09 52 04 | Implement post-editing feature and edit history. Demo shows editing the status text and removing an image. |

PR #934, opened 2023-02-10
IOS-75: Make StatusCardControl an accessibility element

j-f1:


zeitschlag:
@j-f1 Thank you for these improvements. I'd like to hear your opinion, as I tend to postpone this PR until #875 is merged? From my understanding, this PR changes the menu and as we added another menu-entry ("Edit"), we'd have to fix that afterwards (again). That's why I think: Maybe it can wait? What do you think?

j-f1:
Sure, happy to wait and rebase when that PR is merged.

kimar (approved):

PR #936, opened 2023-02-10
IOS-86: Edit L10n.Scene.Favorite.title from "Your Favorites" to "Favorites"

j-f1:
This makes it easier to speak.


Localization/StringsConvertor/input/Base.lproj/app.json
```diff
@@ -620,7 +620,7 @@
             "intro": "These are the posts gaining traction in your corner of Mastodon."
         },
         "favorite": {
-            "title": "Your Favorites"
+            "title": "Favorites"
```
zeitschlag, line 623: Did you change that in Crowdin, too? Otherwise all of those changes to translated `app.json` will be overwritten.

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
zeitschlag: 🙈

Localization/StringsConvertor/input/Base.lproj/app.json
j-f1: Will do!

j-f1: They don’t seem to have a translation for English (regular or US) in Crowdin, but I can monitor after merge to see if the changes get reverted.

zeitschlag: Okay, this seems to be a misunderstanding, sorry for that :) I meant basically every other `app.json` in this PR except the `Base.lproj/app.json`:

Your changes in `da.lproj/app.json` will be overwritten. The change in `he.lproj/app.json`, too, and so on and so forth. Does it make more sense now?

j-f1: Those translations also appear to be blank in the Crowdin UI, e.g. for Hebrew:
Screenshot_2023-02-14 16 14 21

I imagine they’ll update to match the base string (and I don’t want to contribute an English “translation” as that would be wrong)

.vscode/settings.json
zeitschlag: Let me guess: You're using VS Code, right? What is this file good for and why do we need it in the project?

j-f1: We don’t, I just use VS Code as an easy way to edit some files like the `project.pbxproj`. I must be accidentally adding it to the commit — it offers no benefit. The commit below updates the `.gitignore` so it will not happen again.

Localization/StringsConvertor/input/Base.lproj/app.json
zeitschlag: Ah, okay, I think, now I see your point. Took me a second, sorry 🙈 

If there are no translations, the base-strings will be used for those untranslated texts. As you changed the base-string, this will be used and your changes for Hebrew, for example, will be changed from "Favorites" to "Favorites". So in the end you did something, which Crowdin does automatically, but as you changed it to the same outcome, nothing changes :D

PR #937, opened 2023-02-10
IOS-88: Use post content as the primary user input label for posts in feeds/threads

j-f1:
Previously it used the VoiceOver label which starts with the author.


kimar (approved):
Very nice 👍

Issue #940, opened 2023-02-12
App is crashing when writing posts on iPhone


WhereCanI:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When you are writing a new update-adding a post in the iOS app, it will crash once you have written approx. 100-150 characters. First the app gets stuck and then it crashes. 

### Steps To Reproduce

1. Add a new post in the app.
2. Write a story, and once you reach 100-150 characters the app will get stuck and eventually crash.

### Environment

```markdown
- Device: iPhone SE2
- OS: 16.3
- Version:
- Build:
```

j-f1:
Duplicate of https://github.com/mastodon/mastodon-ios/issues/705 — I’ve never been able to reproduce this myself so I haven’t been able to find a real fix but I did open https://github.com/mastodon/mastodon-ios/pull/932 in an attempt to make the issue less severe.

SinghDominik:
I don't know if it is really a duplicate as it literally crashs instead of just hang. And it has nothing to do with an amount of chars because it happens at all char amount ranges.

SinghDominik:
> Duplicate of #705 — I’ve never been able to reproduce this myself so I haven’t been able to find a real fix but I did open #932 in an attempt to make the issue less severe.

I don't know if it is really a duplicate as it literally crashs instead of just hanging. And it has nothing to do with an amount of chars because it happens at all char amount ranges.

Issue #939, opened 2023-02-11
[BUG] Crash after deletion


SinghDominik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I delete a toot and create a new one, Mastodon will crash.

### Steps To Reproduce

1. Delete a toot
2. create a new one

### Environment

```markdown
- Device: iPhone 14
- OS: iOS
- Version: 16.3
```

zeitschlag:
(Most probably) a duplicate of #940

PR #947, opened 2023-02-19
Fix Crash when publishing a new post

zeitschlag:
When posting a new status, the app crashed occasionally when `try save()` in NSManagedObjectContext.swift. It seemed that the instance of NSManagedObjectContext just ... disappeared? The error-message was: `EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)`

It's been pretty tricky to reproduce the buggy behaviour. What worked for me (most of the time) was to write a post, change the language and press "Publish".

(Probably) fixes:

- #940 
- #939
- https://c.im/@Onlineadviser/109831831469594094

### DIFF
--- a/MastodonSDK/Sources/CoreDataStack/Extension/NSManagedObjectContext.swift
+++ b/MastodonSDK/Sources/CoreDataStack/Extension/NSManagedObjectContext.swift
@@ -35,7 +35,7 @@ extension NSManagedObjectContext {
     
     public func performChanges(block: @escaping () -> Void) -> Future<Result<Void, Error>, Never> {
         Future { promise in
-            self.perform {
+            self.performAndWait {
                 block()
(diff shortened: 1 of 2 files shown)


zeitschlag:
@j-f1 I wanted to request a review from you, but Github didn't let me choose you as a reviewer. So, please see this comment as a request and feel free to share you thoughts :) To be honest: You know this piece of code much better and I'm not sure if I broke something with my fix?

j-f1 (approved):
Seems reasonable to me, and I doubt people will notice a brief block on the main thread here.

I’m on 16.4 unable to reproduce because changing the language dismisses the compose view controller. Working on debugging that now.

kimar (approved):
🚀 👍

PR #949, opened 2023-02-21
Hardcoded server url

RuttabHaroon:
search text field hardcoded with url and disabled so that it is not editable. Next button is enabled by default now and server suggestion tableview is hidden


RuttabHaroon:
incorrect gihub repo submission

PR #951, opened 2023-02-23
Improvements to StatusMetricRowView / edit history

j-f1:

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/StatusMetricRowView.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/StatusMetricRowView.swift
@@ -9,6 +9,8 @@ public final class StatusMetricRowView: UIButton {
     let detailLabel: UILabel
     let chevron: UIImageView
 
+    private let contentStack: UIStackView
+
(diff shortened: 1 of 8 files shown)


MastodonSDK/Sources/MastodonUI/View/Content/StatusMetricRowView.swift
```diff
@@ -26,48 +28,92 @@ public final class StatusMetricRowView: UIButton {
...
+        get { super.isHighlighted }
+        set {
+            super.isHighlighted = newValue
+            if newValue {
+                backgroundColor = Asset.Colors.selectionHighlight.color
+            } else {
+                backgroundColor = .clear
+            }
+        }
+    }
```
zeitschlag, line 118: Nice workaround :)

zeitschlag (approved):

MastodonSDK/Sources/MastodonUI/View/Content/StatusMetricRowView.swift
j-f1: Thanks! It’s not 100% because it doesn’t do the nice fade out when you slowly swipe back from the pushed view controller like a real table cell does, but it’s close enough for now.

PR #952, opened 2023-02-23
Fix errors and some warnings when building with Swift 5.8

j-f1:
Warnings with just f268bcd3e: 282
Adding 557c2f870 & df8eb942f: 261
Adding a3b36f710 & 887e619d0: 245

There are a bunch more things that need to be sorted out around Sendable, but hoepfully we can chip away at them slowly (and also hopefully Apple will mark more APIs as Sendable too)

### DIFF
--- a/MastodonSDK/Package.resolved
+++ b/MastodonSDK/Package.resolved
@@ -73,15 +73,6 @@
       }
     },
     {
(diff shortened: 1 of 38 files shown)


kimar (approved):
LGTM 👍  Thank you @j-f1

PR #953, opened 2023-02-24
IOS-70: Improve StatusMetricView Accessibility

j-f1:
- swap from the `UIStackView` back to using manual AutoLayout (it was basically impossible to get the correct behavior out of the stack view, but it wasn’t too bad to do it manually)
- Add a color for the chevron in dark  mode so it isn’t completely invisible
- Make the `StatusMetricView` accessible via VoiceOver


kimar (approved):
🚀  thank you!

PR #954, opened 2023-02-24
Add poll history model to edit post (IOS-2)

kimar:
Unfortunately the data model of poll history diverges from usual poll models, so we need to find a common ground to be able to reuse the existing poll UI to reflect historic data.

Probably the simplest solution would be to extract what we have and make it possible to show poll data for the minimum possible existing model (historic poll data).


Krinkle:
@kimar What does "IOS-2" refer to? I see these in the App Store change log entries as well. Are these something end users can resolve? It appears issue #2 in this iOS repo refers to something else it seems.

j-f1:
They’re listed on https://joinmastodon.org/roadmap.

PR #963, opened 2023-03-03
Don’t assertionFailure() when tapping on non-status cells

j-f1:
I had noticed the app crashing when I touched the activity indicator. This is the reason why. Since it’s an `assertionFailure()` this does not affect production builds. That said, it is annoying since I spend most of my time on a debug build.

I looked at all of the callers of this function and many of them call `assertionFailure` if it returns `nil`. I’ve added `assertionFailure` calls to a couple of places where it makes sense, but now tapping on a non-status cell should no longer cause a crash in debug builds.


kimar (approved):
Makes sense to me 👍  Thank you @j-f1 !

PR #970, opened 2023-03-14
IOS-97: Improve labelling of username field on Create Account screen

kimar:
# Rationale

The username field on create account will now read the full handle and the @ and @domain.tld labels will be hidden from a11y.


j-f1 (approved):
Tested, and it sounds great!

PR #971, opened 2023-03-14
IOS-96: Label Privacy Policy links as links/buttons in the onboarding

kimar:
# Rationale

Label Privacy Policy links as links/buttons in the onboarding.


j-f1 (approved):
Sounds good!

PR #972, opened 2023-03-14
IOS-94: Improve accessibility of the onboarding intro

kimar:
* Fix title of the first page
* Merge accessibility elements for heading and content
* Maybe set an explicit set of accessibilityChildren so that it’s easier to swipe through the onboarding

### DIFF
--- a/Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentCollectionViewCell.swift
+++ b/Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentCollectionViewCell.swift
@@ -68,7 +68,6 @@ class WelcomeContentCollectionViewCell: UICollectionViewCell {
     }
     
     private func updateAccessibility() {
-        accessibilityElements = [titleView, label]
         accessibilityLabel = "\(titleView.accessibilityLabel ?? ""), \(label.accessibilityLabel ?? "")"
         isAccessibilityElement = true
     }
--- a/Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentCollectionViewCell.swift
+++ b/Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentCollectionViewCell.swift
@@ -66,9 +66,17 @@ class WelcomeContentCollectionViewCell: UICollectionViewCell {
         
         NSLayoutConstraint.activate(constraints)
     }
+    
+    private func updateAccessibility() {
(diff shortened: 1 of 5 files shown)


Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentCollectionViewCell.swift
```diff
@@ -66,9 +66,17 @@ class WelcomeContentCollectionViewCell: UICollectionViewCell {
         
         NSLayoutConstraint.activate(constraints)
     }
+    
+    private func updateAccessibility() {
+        accessibilityElements = [titleView, label]
```
j-f1: I believe this is not necessary since setting `isAccessibilityElement` makes the cell its own element with no children (unless you also make it a container).

j-f1 (commented):
Nice, this is a big improvement! There’s one other issue I’m noticing, though: I don’t think I can get to the second tab of the pager view here without using the three-finger scroll gesture. 

The forward (swipe-right) flow goes close button → What is Mastodon? → page control → Create account
The backward (swipe-left) flow goes Create account → page control → How do I pick a server? → close button

I’m not sure how to make this work with a collection view but ideally you’d have the three cells marked as children of the collection view (with `UIAccessibilityContainerType.semanticGroup` on the collection view perhaps?) so that folks can swipe through all three pages and then jump to the action buttons to get started.

Mastodon/Scene/Onboarding/Welcome/View/Pages/WelcomeContentPage.swift
```diff
@@ -49,6 +49,17 @@ enum WelcomeContentPage: CaseIterable {
         }
     }
     
+    var accessibilityLabel: String {
+        switch self {
+        case .whatIsMastodon:
+            return "\(L10n.Scene.Welcome.Education.WhatIsMastodon.title) Mastodon?"
```
zeitschlag: We could use an extra `Localizable.strings`-string here so that translation-folks can include "Mastodon" whereever they want, I'm thinking right-to-left-language

DK1DA: Ah yeah totally! I'll update this 👍

zeitschlag (approved):

kimar:
@j-f1 I've look at the collectionview swipe gesture, but no success so far, what works is focussing the pagecontrol, then doing up/down to go through the pages.

PR #973, opened 2023-03-14
IOS-82: Increase Contrast

kimar:
# Rationale

This PR removes the legacy non-true-dark mode as it has been suboptimal from a contrast perspective and didn't reflect the current branding in an accurate way.


MastodonSDK/Sources/MastodonCore/Service/Theme/ThemeService.swift
```diff
@@ -20,7 +20,7 @@ public final class ThemeService {
     public let currentTheme: CurrentValueSubject<Theme, Never>
 
     private init() {
-        let theme = ThemeName(rawValue: UserDefaults.shared.currentThemeNameRawValue)?.theme ?? ThemeName.mastodon.theme
+        let theme = ThemeName.system.theme
```
zeitschlag, line 23: If there's only one theme left... do we still need the `ThemeService`? Or is it just way too much work to get rid of it right away?

MastodonSDK/Sources/MastodonLocalization/Resources/Base.lproj/Localizable.strings
```diff
@@ -502,7 +502,6 @@ uploaded to Mastodon.";
 "Scene.Settings.Section.Preference.DisableEmojiAnimation" = "Disable animated emojis";
 "Scene.Settings.Section.Preference.OpenLinksInMastodon" = "Open links in Mastodon";
 "Scene.Settings.Section.Preference.Title" = "Preferences";
-"Scene.Settings.Section.Preference.TrueBlackDarkMode" = "True black dark mode";
```
zeitschlag, line 505: Did you remove `true_black_dark_mode` from `Localization/app.json`, too?

zeitschlag (approved):
LGTM 👍 Maybe double-check `Localization/app.json`

MastodonSDK/Sources/MastodonCore/Service/Theme/ThemeService.swift
DK1DA: I thought that it might be handy if we, at some point, maybe decide to bring back different themes, not sure if that's worth it though.

MastodonSDK/Sources/MastodonLocalization/Resources/Base.lproj/Localizable.strings
DK1DA: Yes, it's here: https://github.com/mastodon/mastodon-ios/pull/973/files#diff-527687b68684efcd790236be5e68527ce4037772106c748a11729b8538367c20L725

PR #975, opened 2023-03-14
IOS-69: Label this “re-blog” otherwise it sounds like “reh-blog”

kimar:
# Rationale

The reblog button in `ActionToolbarContainer` is read as "reh-blog" instead of "re-blog" which sounds a bit odd in english. I've chosen to introduce a localizable key purely used for A11y to make this sound nicer.

# Warning

We need to make sure to pull new strings in before releasing due to the updated `Localizable.stringsdict` with new pluralisations.

### DIFF
--- a/Localization/Localizable.stringsdict
+++ b/Localization/Localizable.stringsdict
@@ -231,6 +231,28 @@
 			<string>%ld reblogs</string>
 		</dict>
 	</dict>
+		<key>plural.count.reblog_a11y</key>
+    	<dict>
+    		<key>NSStringLocalizedFormatKey</key>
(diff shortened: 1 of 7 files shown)


j-f1 (approved):
Nice!

j-f1:
Based on bullet 4 under “Home and Posts” in IOS-113 I think this could also be fixed by relabeling it “boost” across the app

kimar:
@j-f1 good point! ~Hm, I'll change the i18n inside the localized files then, but this also needs to be taken upstream to Crowdin~ Let's keep it for the VQA updates, I don't want to mix too many different topics into this PR.

PR #976, opened 2023-03-14
IOS-110: Fix Cannot switch accounts when large content viewer is enabled

kimar:
# Rationale

when enabling large text, and further increasing size to use the large content viewer, it's long-press gesture interferes with the long-press for the account switcher.

This PR allows simultaneous gestures so the large content HUD shows up and the account switcher will open up as well.


zeitschlag (approved):
Very nice!

PR #977, opened 2023-03-14
IOS-109: Fix Large content viewer does not appear when pressing and holding the Compose button

kimar:
# Rationale

Added `UILargeContentViewerInteraction` to compose button.

### DIFF
--- a/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
+++ b/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
@@ -370,6 +370,9 @@ extension MainTabBarController {
         
         updateTabBarDisplay()
         
+        composeButton.addInteraction(UILargeContentViewerInteraction())
+        composeButton.showsLargeContentViewer = true
+        composeButton.largeContentTitle = L10n.Common.Controls.Actions.compose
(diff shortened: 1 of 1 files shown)


j-f1:
If bullet 4 from IOS-113 is going to be implemented I believe this will not be necessary.

Otherwise, there’s a small issue with the current implementation: while you can press on any of the regular tab icons and then drag across to view the other tabs, dragging off of the compose button dismisses the large content viewer and you must press again to view the other tabs.

kimar:
@j-f1 I've looked into the issue and tried out a couple of things now, I think the issue correlates with the touch up outside cancellation of the compose `UIButton`, I haven't found a fix / workaround for this yet, but given IOS-113 bullet 4 is done this get's obsolete (as you already mentioned). 

Now the question is: Shall we merge it or discard it? Any opinion @zeitschlag ?

zeitschlag:
I see the issue with dragging across different icons (I didn't know that that was possible, pretty cool!), and I didn't have neither hard feelings nor an opinion there. 

But I tended to discarding this PR in favor of replacing the Compose-button with the new design (IOS-113, bullet 4) right away. I wanted to give it a try and so I did in #980

kimar:
@zeitschlag yes I totally agree! Let's close this one in favor of yours!

PR #979, opened 2023-03-15
IOS-80 Add AXCustomContent to ProfileCardView

j-f1:
I’m not super happy with how the mutuals one reads “foo, bar, and 2 other mutual mutuals” but ¯\\\_(ツ)\_/¯


kimar (approved):
LGTM 🚀

PR #982, opened 2023-03-19
fix spelling in doc

Lawhacknifemi:
entitles should be entities


kimar (approved):
LGTM 🚀  Thank you @Lawhacknifemi

PR #989, opened 2023-03-22
IOS-129: Redesign onboarding buttons and add loading indicator on Next-Button

kimar:
# Rationale

Redesign onboarding buttons and add loading indicator on Next-Button to match current designs.

I have reformatted `MastodonLoginViewController` to match the default indentation, so there are many changes show, sorry for that :-/

# Screenshots

| Before | After |
|---|---|
|Simulator Screen Shot - iPhone 14 Pro - 2023-03-21 at 08 46 48|Simulator Screen Shot - iPhone 11 Pro Max - 2023-03-22 at 20 39 56|

### DIFF
--- a/Mastodon/Scene/Onboarding/Login/MastodonLoginViewController.swift
+++ b/Mastodon/Scene/Onboarding/Login/MastodonLoginViewController.swift
@@ -14,240 +14,239 @@ import AuthenticationServices
 import MastodonLocalization
 
 protocol MastodonLoginViewControllerDelegate: AnyObject {
-  func backButtonPressed(_ viewController: MastodonLoginViewController)
-  func nextButtonPressed(_ viewController: MastodonLoginViewController)
+    func backButtonPressed(_ viewController: MastodonLoginViewController)
+    func nextButtonPressed(_ viewController: MastodonLoginViewController)
 }
 
 enum MastodonLoginViewSection: Hashable {
-  case servers
+    case servers
 }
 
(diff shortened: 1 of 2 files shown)


Mastodon/Scene/Onboarding/Login/MastodonLoginViewController.swift
```diff
@@ -14,279 +14,305 @@ import AuthenticationServices
...
-    guard let userInfo = notification.userInfo,
-          let duration = userInfo[UIResponder.keyboardAnimationDurationUserInfoKey] as? NSNumber
-    else { return }
-
-    contentView.bottomConstraint?.constant = 0
+}
 
-    UIView.animate(withDuration: duration.doubleValue, delay: 0, options: .curveEaseInOut) {
-      self.view.layoutIfNeeded()
+private extension MastodonLoginViewController {
```
zeitschlag: No need to change, but a question: What's the reason to put `setRightBarButtonState(_:)` into its own extension?

zeitschlag (approved):
Looks good!

Mastodon/Scene/Onboarding/Login/MastodonLoginViewController.swift
DK1DA: Tbh, no real reason, I'll move it to the class directly. I actually like to split things e.g. all things private. But given this is the only function I don't see a strong reason to do this.

PR #991, opened 2023-03-23
IOS-131: Remove blurry background, make pageControl background promintent, remove inset

kimar:
# ToDos

- [x] Welcome screen page dots are obscured by artwork. Set a prominent background as always
- [x] Text should not be on a blur view, should be closer to the top (roughly 80pt)
- [x] Create account button should be title case
  * I updated the string for `en`  in Crowdin


zeitschlag (approved):
The code looks fine to me. I like, that you fixed the background-style of the page-control!

------

Some background-information (pun intended): The reason I added a blurry background back then was, that on smaller screens like iPhone SE, the text is not readable as the plane crashes into it:

PR #992, opened 2023-03-27
Port the media layout algorithm from the Android app

grishka:
As discussed, I tried to port the media layout algorithm from the Android app. Android sources: [the layout algorithm itself](https://github.com/mastodon/mastodon-android/blob/7b0a3f0f96af27aac39c60a2ff095472656ea689/mastodon/src/main/java/org/joinmastodon/android/ui/PhotoLayoutHelper.java) and [the custom ViewGroup](https://github.com/mastodon/mastodon-android/blob/7b0a3f0f96af27aac39c60a2ff095472656ea689/mastodon/src/main/java/org/joinmastodon/android/ui/views/MediaGridLayout.java) that actually puts the views on the screen. This new layout algorithm works with up to 10 media attachments.

IMG_0876

I'm not familiar with Swift and UIKit, and definitely not with ReactiveSwift that permeates the entire codebase of the app. I might have done something stupid.

This PR isn't quite ready for merging. There's something inside `MediaView` that doesn't respect its frame and won't crop vertically, resulting in overlapping views:
IMG_0874
IMG_0875

There's also a bug I don't know how to fix because I don't know where to hook into either the app or UIKit to re-run the layout with new `maxSize` when the view size changes. Android would automatically measure and then lay out the view subtree when the dimensions of its parent `ViewGroup` change, but iOS apparently doesn't do that? To reproduce, open an attachment, rotate the device to landscape orientation, then close the attachment.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
+++ b/MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
@@ -30,7 +30,7 @@ class MediaLayoutHelper {
 	static let maxHeight: Float = 1777
 	static let minHeight: Float = 563
 	static let gap: Float = 1.5
-	static let maxRatio: Float = Float(maxWidth) / Float(maxHeight)
+	static let maxRatio = maxWidth / maxHeight
 	
 	public static func generateMediaLayout(attachments: [MastodonAttachment]) -> MediaLayoutResult? {
 		if attachments.count<2 {
@@ -42,7 +42,7 @@ class MediaLayoutHelper {
 		var allAreSquare = true
 		for att in attachments {
 			let ratio: Float = Float(att.size.width/att.size.height)
-			if ratio<=1.2 {
+			if ratio <= 1.2 {
 				allAreWide = false
 				if ratio<0.8 {
 					allAreSquare = false
@@ -282,7 +282,7 @@ class MediaLayoutHelper {
 				columnSizes.append(offset - gridLineOffsets[i]) // i is already offset by one here
 			}
 			
-			for (row, _) in rowTiles.enumerated() {
+			for row in 0..<rowTiles.count {
 				var columnOffset: Int = 0
 				for (tile, _) in rowTiles[row].enumerated() {
 					let startColumn = columnOffset
--- a/MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
+++ b/MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
@@ -10,305 +10,305 @@ import MastodonSDK
 import CoreDataStack
 
 public struct MediaLayoutResult {
-	let width: Int
-	let height: Int
-	let columnSizes: [Int]
-	let rowSizes: [Int]
-	let tiles: [Tile]
-	
-	public struct Tile {
-		var colSpan: Int
-		let rowSpan: Int
-		var startCol: Int
-		let startRow: Int
-		var width: Int = 0
-	}
+    let width: Int
+    let height: Int
+    let columnSizes: [Int]
+    let rowSizes: [Int]
+    let tiles: [Tile]
+    
+    public struct Tile {
+        var colSpan: Int
+        let rowSpan: Int
+        var startCol: Int
+        let startRow: Int
+        var width: Int = 0
+    }
 }
 
 class MediaLayoutHelper {
-	static let maxWidth: Float = 1000
-	static let maxHeight: Float = 1777
-	static let minHeight: Float = 563
-	static let gap: Float = 1.5
-	static let maxRatio = maxWidth / maxHeight
-	
-	public static func generateMediaLayout(attachments: [MastodonAttachment]) -> MediaLayoutResult? {
-		if attachments.count<2 {
-			return nil
-		}
-		
-		var ratios: [Float] = []
-		var allAreWide = true
-		var allAreSquare = true
-		for att in attachments {
-			let ratio: Float = Float(att.size.width/att.size.height)
-			if ratio <= 1.2 {
-				allAreWide = false
-				if ratio<0.8 {
-					allAreSquare = false
-				}
-			} else {
-				allAreSquare = false
-			}
-			ratios.append(ratio)
-		}
-		
-		let avgRatio: Float = ratios.reduce(0.0, +) / Float(ratios.count)
-		
-		switch attachments.count {
-		case 2:
-			if allAreWide && avgRatio>1.4*maxRatio && (ratios[1]-ratios[0])<0.2 {
-				// Two wide attachments, one above the other
-				let h = Int(max(min(maxWidth/ratios[0], min(maxWidth/ratios[1], (maxHeight-gap)/2.0)), minHeight/2.0).rounded())
-				
-				return MediaLayoutResult(width: Int(maxWidth),
-								  height: Int((Float(h)*2.0+gap).rounded()),
-								  columnSizes: [Int(maxWidth)],
-								  rowSizes: [h, h],
-								  tiles: [
-									MediaLayoutResult.Tile(colSpan: 1, rowSpan: 1, startCol: 0, startRow: 0),
-									MediaLayoutResult.Tile(colSpan: 1, rowSpan: 1, startCol: 0, startRow: 1)
-								  ])
-			} else if allAreWide || allAreSquare {
-				// Next to each other, same ratio
-				let w: Float = (maxWidth-gap) / 2.0
-				let h: Float = max(min(w/ratios[0], min(w/ratios[1], maxHeight)), minHeight)
-				
-				let wInt: Int = Int(w.rounded())
-				let hInt: Int = Int(h.rounded())
-				
-				return MediaLayoutResult(width: Int(maxWidth),
-										 height: hInt,
-										 columnSizes: [wInt, Int(maxWidth)-wInt],
-										 rowSizes: [hInt],
-										 tiles: [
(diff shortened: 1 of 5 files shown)


grishka:
Also, teeny-tiny gifs. I'm definitely missing some crucial piece of understanding of how UIKit views work.

IMG_0877

j-f1:
I think you need to call `super.layoutSubviews()` in your `layoutSubviews` otherwise the subviews won’t layout their own subviews.

MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
```diff
@@ -0,0 +1,314 @@
...
+		var colSpan: Int
+		let rowSpan: Int
+		var startCol: Int
+		let startRow: Int
+		var width: Int = 0
+	}
+}
+
+class MediaLayoutHelper {
+	static let maxWidth: Float = 1000
```
j-f1: Most of the rest of the code uses `CGFloat` (aka double) because that’s what the UIKit layout APIs accept.

MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
```diff
@@ -0,0 +1,314 @@
...
+	static let maxHeight: Float = 1777
+	static let minHeight: Float = 563
+	static let gap: Float = 1.5
+	static let maxRatio: Float = Float(maxWidth) / Float(maxHeight)
```
j-f1: In this case you don’t need any type annotations or explicit casts:
```suggestion
	static let maxRatio = maxWidth / maxHeight
```

grishka:
Nope, it made no difference =/

MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
```diff
@@ -0,0 +1,314 @@
...
+		if attachments.count<2 {
+			return nil
+		}
+		
+		var ratios: [Float] = []
+		var allAreWide = true
+		var allAreSquare = true
+		for att in attachments {
+			let ratio: Float = Float(att.size.width/att.size.height)
+			if ratio<=1.2 {
```
j-f1: Can you add spaces around the operators? The rest of the app (and most Swift code except for the `..<` operator) do that and it improves readability a lot IMO
```suggestion
			if ratio <= 1.2 {
```

MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
```diff
@@ -0,0 +1,314 @@
...
+										 columnSizes: [Int(w0.rounded()), Int(w1.rounded()), Int(maxWidth-w0.rounded()-w1.rounded())],
+										 rowSizes: [Int(hCover.rounded()), Int(h.rounded())],
+										 tiles: [
+											MediaLayoutResult.Tile(colSpan: 3, rowSpan: 1, startCol: 0, startRow: 0),
+											MediaLayoutResult.Tile(colSpan: 1, rowSpan: 1, startCol: 0, startRow: 1),
+											MediaLayoutResult.Tile(colSpan: 1, rowSpan: 1, startCol: 1, startRow: 1),
+											MediaLayoutResult.Tile(colSpan: 1, rowSpan: 1, startCol: 2, startRow: 1)
+										 ])
+			} else {
+				// One on the left, three smaller ones on the right
```
j-f1: Does this mean 4 attachments will never be rendered as a 2x2 grid?

MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
```diff
@@ -0,0 +1,314 @@
...
+			
+			gridLineOffsets = gridLineOffsets.sorted()
+			gridLineOffsets.append(Int(maxWidth))
+			
+			var columnSizes: [Int] = [gridLineOffsets[0]]
+			for (i, offset) in gridLineOffsets[1...].enumerated() {
+				columnSizes.append(offset - gridLineOffsets[i]) // i is already offset by one here
+			}
+			
+			for (row, _) in rowTiles.enumerated() {
```
j-f1: I don’t think you need the `enumerated` here. If you want the values:

```suggestion
			for row in rowTiles {
```
 or if you want the indices:

```suggestion
			for row in 0..<rowTiles.count {
```

MastodonSDK/Sources/MastodonUI/View/Container/MediaGridContainerView.swift
```diff
@@ -183,11 +183,12 @@ extension MediaGridContainerView {
...
         let maxSize: CGSize
+		let layout: MediaLayoutResult
         
-        init(count: Int, maxSize: CGSize) {
-            self.count = min(count, 9)
+		init(count: Int, maxSize: CGSize, layout: MediaLayoutResult) {
+            self.count = min(count, 10)
             self.maxSize = maxSize
-        
+			self.layout = layout
```
j-f1: Looks like indentation got a little misaligned here? In Xcode, Editor → Structure → Re-Indent should match most indentation to the rest of the project.

MastodonSDK/Sources/MastodonUI/View/Container/MediaGridContainerView.swift
```diff
@@ -200,74 +201,96 @@ extension MediaGridContainerView {
...
-                    bottomHorizontalStackView.addArrangedSubview(mediaViewOrPlaceholderView(at: 8))
-                }
-            default:
-                assertionFailure()
-                return
-            }
+			
+			if count<2 || count>maxCount {
+				assertionFailure("unexpected attachment count \(count)")
+				return
```
j-f1: Since `assert` APIs including `assertionFailure` do nothing in prod builds it would be great to have some sensible behavior here. (If truly impossible this should probably be a `fatalError` or `precondition` instead)

MastodonSDK/Sources/MastodonUI/View/Container/MediaGridContainerView.swift
```diff
@@ -200,74 +201,96 @@ extension MediaGridContainerView {
...
                 view.widthAnchor.constraint(equalToConstant: containerWidth).priority(.required - 1),
                 view.heightAnchor.constraint(equalToConstant: containerHeight).priority(.required - 1),
             ])
         }
     }
 }
+
+class GridLayoutView : UIView {
+	private var layout: MediaLayoutResult?
+	private var measuredHeight = 0
```
j-f1: You can use `private(set)` to make the property readonly to external code without needing a getter:
```suggestion
	private(set) var measuredHeight = 0
```

MastodonSDK/Sources/MastodonUI/Helper/MediaLayoutHelper.swift
grishka: Yes, it's always 1 bigger one and 3 smaller ones

MastodonSDK/Sources/MastodonUI/View/Container/MediaGridContainerView.swift
grishka: This is a sanity check — this condition should really not be possible because it's checked elsewhere. `AdaptiveLayout` is used when there's only one attachment, nothing when zero, and attachments beyond max count are discarded. In Java I'd `throw new IllegalStateException(...)` here which would crash the app.

j-f1: That makes sense — that would be `fatalError` in Swift then.

Or:

```swift
precondition(count >= 2 && count <= maxCount, "Unexpected attachment count \(count)")
```

MastodonSDK/Sources/MastodonUI/View/Container/MediaGridContainerView.swift
grishka: I think I found a bug in xcode. I have my global settings set to indent with tabs. It re-indents with tabs even when I set spaces in "text settings" on the right panel for that particular file.

grishka:
I limited the aspect ratio to 9:20 to avoid weird skinny layouts with images of extreme aspect ratios. I chose 9:20 specifically so that vertical screenshots from modern phones don't end up getting cropped. Modern "bezel-less" phones have 9:18-ish screens.

Simulator Screen Shot - iPhone 14 Pro - 2023-03-30 at 18 56 43

grishka:
Another interesting observation — if a post has a broken layout, but I then tap it to open separately, *that* screen has a correct layout.

Might it be that images loading asynchronously affect the layout inside MediaView somehow?

grishka:
I decided to take another look at it and remembered that the layout inspector exists. So something is adding these constraints to the UIImageView, making the layout break when I set frames manually. Now it's only a matter of finding what it is.
Снимок экрана 2024-03-30 в 12 50 12

tvrrp:
@grishka 
Constraints inside MediaView should not affect the size of the MediaView itself.

The behavior where the image extends beyond the bounds occurs because MediaView instances are reused in `statusView.mediaGridContainerView.dequeueMediaView`. When using AdaptiveLayout, translatesAutoresizingMaskIntoConstraints is disabled, but it is not re-enabled when using GridLayout. This likely causes UIKit to calculate the size of MediaView based on the image dimensions.

You can add `view.translatesAutoresizingMaskIntoConstraints = true` in `MediaGridContainerView.prepareForReuse()`. Or, use frame-based layout for AdaptiveLayout as well.

grishka:
Yeah, I figured that out. The problem was that, iirc, when there was a single attachment, `view.translatesAutoresizingMaskIntoConstraints = true` was being set on it, which I didn't clear, which broke my manual setting of view frames next time when that view got reused. It all works now.

PR #994, opened 2023-03-27
IOS-132: Server picker VQA improvements

kimar:
# Rationale

- [x] Should be an inline nav bar & “Pick server“ should be title case
- [x] Chips should be larger — increase padding
- [ ] Remove stroke on selected chip in light mode.
  - I'm not seeing any stroke there currently
- [x] “Next” button should be bold
- [x] Language picker
  - [x] What is “be” and why does it not follow suit with other languages
    - Needs to be fixed with response of https://api.joinmastodon.org/languages
- [x] Picking categories doesn’t scroll you all the way to the top of the list, cuts off the first server in the category
- [ ] Full back swipe when there should only be edge back swipe on this screen
  - Tbd as this is somewhat a custom global behavior
- [x] No touch down state on server cells, can’t tell you’re tapping on a server until you let go

### DIFF
--- a/Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewController.swift
+++ b/Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewController.swift
@@ -141,6 +141,13 @@ extension MastodonPickServerViewController {
             )
         }
         .store(in: &disposeBag)
+        
+        viewModel.scrollToTop
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.tableView.scroll(to: .top, animated: false)
+            }
+            .store(in: &disposeBag)
 
         authenticationViewModel
             .authenticated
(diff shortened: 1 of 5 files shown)


Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewController.swift
```diff
@@ -371,6 +378,13 @@ extension MastodonPickServerViewController: UITableViewDelegate {
         guard case let .server(server, _) = item else { return }
         tableView.selectRow(at: indexPath, animated: false, scrollPosition: .none)
         viewModel.selectedServer.send(server)
+        
+        // Briefly highlight selected cell
```
zeitschlag, line 382: Uh, I like that behaviour!

Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewController.swift
```diff
@@ -141,6 +141,13 @@ extension MastodonPickServerViewController {
             )
         }
         .store(in: &disposeBag)
+        
+        viewModel.scrollToTop
```
zeitschlag, line 145: This kind of feels ... weird for something as simple as scrolling to the top. Usually I'd have used a method for this, but I still live in the old school object oriented world :D

zeitschlag (approved):
Nice!

PR #998, opened 2023-03-29
Make UITableView background color clear IOS-137

kimar:
In Onboarding -> Privacy

- [x] “I agree“ should be title case
- [x] bBackground on table group should be unsetxa


zeitschlag (approved):
Nice!

PR #999, opened 2023-03-31
Improve Create Account VQA IOS-138

kimar:
## Create account

- [x] Title should be title case
- [x] Bottom of the screen has a black section
- [x] As soon as you start typing one character into the password field, both fields turn red
- [ ] Eight character password requirement isn’t in the designs
- [x] If you enter a username that’s taken, the error only shows up when you tap out of the field & disappears when you tap back in (probably should stay until you start typing to edit)
- [x] Copy is incorrect ([see design](https://www.figma.com/file/vWXrZ38w10dyH7ZiKGN7eA/Onboarding?node-id=10%3A8147&t=U0QLYVf6aujNTuNk-4))


kimar:
Not sure what to do with the 8-character password criteria @Gargron I've left this as-is for now, not sure if it needs to be changed in the app or the design documents, but I didn't want to mess around with our password criteria just now.

zeitschlag (approved):
Once again: Nice 🎉

PR #1000, opened 2023-03-31
Improve style of "Check your Inbox" screen IOS-139

kimar:
- [x] “Link“ shouldn’t be capitalized
- [x] Use smart apostrophe in “Didn’t“
- [x] Numbers should be monospaced
- [x]  “Check your email“ alert should be title case
- [ ] This alert needs some thought—do we actually want an alert here?


kimar:
No action taken on the last bullet, I think this is a conceptual / design decision which we should take.

zeitschlag (approved):
Nice 🎉

PR #1012, opened 2023-04-13
Put ComposeToolbar-actions in ScrollView

### LINKED ISSUE #1009: [BUG] Accessibility — display size

### Current Behavior

I can't see how many characters I have left to type

### Expected Behavior

I see a number

### Steps To Reproduce

1. Open settings 
2. Display
3. Display zoom
4. Larger text
5. Open Mastodon

### Environment

```markdown
- Device: iPhone SE
- OS: iOS 16.4.1
- Version: v1.5.1
- Build: 303
```

### Anything else?

_No response_

zeitschlag:
Fixes #1009.

The buttons are in a `ScrollView`, but scrolling is only enabled if zoom is enabled.
Speaking of which: This is how you change it in the simulator:

<video src="https://user-images.githubusercontent.com/2580019/231877844-ddaf719c-875b-4d80-bf3f-b392115e2c53.mp4"></video>

And this is how the result looks like for enabled zoom:

<video src="https://user-images.githubusercontent.com/2580019/231877920-84593914-30ac-474d-ae28-89a233be7142.mp4"></video>

For Default-Zoom nothing has changed, scrolling is disabled here:

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/ComposeContentToolbarView.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/ComposeContentToolbarView.swift
@@ -18,8 +18,6 @@ protocol ComposeContentToolbarViewDelegate: AnyObject {
 
 struct ComposeContentToolbarView: View {
     
-    let logger = Logger(subsystem: "ComposeContentToolbarView", category: "View")
-    
     static var toolbarHeight: CGFloat { 48 }
     
     @ObservedObject var viewModel: ViewModel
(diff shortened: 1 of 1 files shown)


j-f1 (approved):
Nice improvement! I noticed that the actions also overflow on 4-inch screens (SE 1 & iPod touch), but since those devices don’t run iOS 16 I’m not sure it’s worth worrying about them.

A future optimization for this could be to switch to something like this once we do drop iOS 15 support:

```swift
var toolbarContent: some View {
  HStack(spacing: 0) { ... }
}
var body: some View {
  ViewThatFits {
    toolbarContent
    ScrollView(.horizontal) {
      toolbarContent
    }
  }
}
```

(You could apply this change with the current behavior as a fallback for iOS 15 but I’m not sure if that’s worth the dev time)

kimar (approved):
Very nice 👍  LGTM! 🚀

PR #1018, opened 2023-04-15
Fix new crash when closing compose VC

j-f1:
It seems that when dismissing the compose view controller, the system (sometimes? always? not sure) goes through this code path after `view.window` has been set to `nil`. I think it’s safe to skip the update in this case to avoid the crash.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewController.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewController.swift
@@ -207,8 +207,9 @@ extension ComposeContentViewController {
             self.tableView.contentInset.bottom = padding - self.view.safeAreaInsets.bottom
             self.tableView.verticalScrollIndicatorInsets.bottom = padding - self.view.safeAreaInsets.bottom
             UIView.animate(withDuration: 0.3) {
+                guard let window = self.view.window else { return }
                 // ref: https://developer.apple.com/documentation/uikit/uiresponder/1621578-keyboardframeenduserinfokey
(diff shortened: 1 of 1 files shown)


MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewController.swift
```diff
@@ -207,8 +207,9 @@ extension ComposeContentViewController {
             self.tableView.contentInset.bottom = padding - self.view.safeAreaInsets.bottom
             self.tableView.verticalScrollIndicatorInsets.bottom = padding - self.view.safeAreaInsets.bottom
             UIView.animate(withDuration: 0.3) {
+                guard let window = self.view.window else { return }
                 // ref: https://developer.apple.com/documentation/uikit/uiresponder/1621578-keyboardframeenduserinfokey
-                let localKeyboardFrame = self.view.convert(endFrame, from: self.view.window!.screen.coordinateSpace)
+                let localKeyboardFrame = self.view.convert(endFrame, from: window.screen.coordinateSpace)
```
zeitschlag, line 212: Great, that you found that! What I didn't understand Why are we using `window.screen.coordinateSpace` here and not `view.coordinateSpace`? But then I ran the app on the iPad Simulator and found the answer.

zeitschlag (approved):

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewController.swift
j-f1: Yep!

PR #1024, opened 2023-04-21
Fix “hide sensitive content” button being hidden

j-f1:
Previously, tapping on the 👁️ button would hide it, leaving no way to hide nsfw/content warning content after showing it.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
@@ -311,14 +311,15 @@ extension StatusView.ViewModel {
                 // statusView.spoilerBannerView.label.reset()
             }
             
-            if statusView.style == .editHistory, let spoilerContent = spoilerContent, !spoilerContent.string.isEmpty {
+            if statusView.style == .editHistory {
+                statusView.setContentSensitiveeToggleButtonDisplay(isDisplay: false)
+            }
+            if let spoilerContent = spoilerContent, !spoilerContent.string.isEmpty {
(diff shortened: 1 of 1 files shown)


zeitschlag (changes requested):
@j-f1 Great! Thank you for giving this a try, I really appreciate it 👍 

From my perspective, something's a bit odd with the visibilty of the 👁️ -button. Take this video, for example:

<video src="https://user-images.githubusercontent.com/2580019/233733807-bf49536f-b73d-4973-96c8-a913d2b9214d.mp4"></video>

When I see a CW-d post and want to make it visible, I'd expect the crossed eye to appear instead so that I can hide the content again. But on the main screen in the video, it's not. It appears if I'd scroll and scroll back, so somethingsomething `reuseCell` and `cellForForAtIndexPath`, I'd think?

Also on the detail-page it feels ... weird. I don't get it when the crossed 👁️ is there and when it's not? Does it make sense to you what I mean? If you have any question, just hit me up, I'm more than happy to help out.

Maybe you can have another look at this issue? That'd be awesome 🥳

j-f1:
Are you sure you’re running on this branch? That describes the behavior I was seeing before making this change.

zeitschlag:
I thought so, yeah, but to be sure I will give it another try later, okay? Sorry for the confusion 😵‍💫

zeitschlag (approved):

zeitschlag:
@j-f1 So, I checked again and you're absolutely right: It works 👍 

I'm really sorry again for the confusion and wrong review. I have no idea what I did differently yesterday, I double-checked before posting the prior comment, but something must have slipped.

Thank you for fixing this bug!

PR #1032, opened 2023-05-04
IOS-140: Implement Verified Links on UserView

kimar:
# Rationale

This PR introduces verified links on the UserView. Also there is a button which can be used to follow/unfollow/block users quickly without having to view their profiles.

# Acceptance Criteria

Preview and Context Menu have been moved into IOS-156

- [x] If we can't show more than 3 characters of the username, just hide it and give the remaining space to the display name.
- [x] The Follow button should be fixed width so we don't create layout shift when we switch from "Follow" to "Following". I'm not hellbent on 88 if there's a smarter way to do this.
- [x] Switch to decimal abbreviations after 999 followers (1.1K, 10.2K)
- [x] We're trying to avoid shaming people without verification. If the user does not have an associated link, we use the passive voice ("No verified link" and a question mark symbol).

# Demo

| Light Appearance | Dark Appearance |
|---|---|
| Simulator Screenshot - iPhone 14 Pro - 2023-05-04 at 11 55 58 | Simulator Screenshot - iPhone 14 Pro - 2023-05-04 at 11 55 51 |


Mastodon/Diffable/Search/SearchHistorySection.swift
```diff
@@ -28,7 +28,7 @@ extension SearchHistorySection {
         let userCellRegister = UICollectionView.CellRegistration<SearchHistoryUserCollectionViewCell, ManagedObjectRecord<MastodonUser>> { cell, indexPath, item in
             context.managedObjectContext.performAndWait {
                 guard let user = item.object(in: context.managedObjectContext) else { return }
-                cell.configure(viewModel: .init(value: user))
+                cell.configure(viewModel: .init(value: user), delegate: configuration.searchHistorySectionHeaderCollectionReusableViewDelegate)
```
zeitschlag: This is a very java-esque, but awesome property name :D Made me smile!

Mastodon/Scene/Profile/FamiliarFollowers/FamiliarFollowersViewController.swift
```diff
@@ -91,4 +93,15 @@ extension FamiliarFollowersViewController: UITableViewDelegate, AutoGenerateTabl
 }
 
 // MARK: - UserTableViewCellDelegate
-extension FamiliarFollowersViewController: UserTableViewCellDelegate { }
+extension FamiliarFollowersViewController: UserTableViewCellDelegate {
```
zeitschlag: Where do we use the `FamiliarFollowersViewController`?

MastodonSDK/Sources/MastodonUI/View/Content/UserView+ViewModel.swift
```diff
@@ -22,10 +26,17 @@ extension UserView {
...
+        @Published public var authorFollowers: Int?
+        @Published public var authorVerifiedLink: String?
+        @Published public var user: MastodonUser?
     }
 }
 
 extension UserView.ViewModel {
+    private static var metricFormatter: MastodonMetricFormatter = {
+        return MastodonMetricFormatter()
+    }()
```
zeitschlag: Is there are a reason that this is a computed property and not just something like 

```suggestion
    private static var metricFormatter = MastodonMetricFormatter()
```

? Or is it merely to match the coding-style?

MastodonSDK/Sources/MastodonUI/Extension/NSAttributedString+Format.swift
```diff
@@ -0,0 +1,16 @@
+// Copyright © 2023 Mastodon gGmbH. All rights reserved.
+
+import Foundation
+
+public extension NSAttributedString {
+    convenience init(format: NSAttributedString, args: NSAttributedString...) {
+        let mutableNSAttributedString = NSMutableAttributedString(attributedString: format)
+
+        args.forEach { attributedString in
+            let range = NSString(string: mutableNSAttributedString.string).range(of: "%@")
```
zeitschlag: Neat extension! 
Does this also work in case of multiple `%@` in the `format`-String?

MastodonSDK/Sources/MastodonUI/View/Content/UserView+ViewModel.swift
```diff
@@ -74,5 +85,43 @@ extension UserView.ViewModel {
...
+            .sink { link in
+                userView.authorVerifiedImageView.image = link == nil ? UIImage(systemName: "questionmark.circle") : UIImage(systemName: "checkmark")
+
+                switch link {
+                case let .some(link):
+                    userView.authorVerifiedImageView.tintColor = Asset.Colors.brand.color
+                    userView.authorVerifiedLabel.textColor = Asset.Colors.brand.color
+                    do {
+                        let mastodonContent = MastodonContent(content: link, emojis: [:])
+                        let content = try MastodonMetaContent.convert(document: mastodonContent)
```
zeitschlag, line 110: Great, that you consider emojis etc. in the account-name. That's something I'm pretty sured I'd have missed!

MastodonSDK/Sources/MastodonUI/View/Content/UserView.swift
```diff
@@ -95,3 +214,78 @@ extension UserView {
     }
     
 }
+
+private final class FollowButton: RoundedEdgesButton {
```
zeitschlag, line 219: Just a thought: We could think about porting `RoundedEdgesButton` to use `UIButton.Configuration`

MastodonSDK/Sources/MastodonUI/View/Content/UserView.swift
```diff
@@ -38,13 +54,75 @@ public final class UserView: UIView {
...
+    }()
+    
+    private let verifiedStackCenterSpacerView: UILabel = {
+        let label = UILabel()
+        label.text = " · "
+        label.textColor = .secondaryLabel
+        return label
+    }()
+    
+    private let followButton: UIButton = {
```
zeitschlag:
```suggestion
    private let followButton: FollowButton = {
```

zeitschlag (commented):
Thank you @kimar, this looks good, I like it!

Apart from the code I noticed two things:

1. Recent Searches

The (verified) link collides with the number of followers:



On other screens, the link is below the number of followers. I'm pretty sure @samhenrigold has thoughts/ideas regarding this, too?

2. FollowButton-state on other profiles

This feels more of a bug and I have no idea where this comes from:
Being `@nathan@berlin.social`, I went to `@zeitschlag@chaos.social`'s Profile-Screen and then I tapped the "Following"-button. There I see the people from `berlin.social` whom `zeitschlag` follows but as `@nathan@berlin.social` I follow none of them. Nevertheless the button states "Following":



When I tap the "Following"-button although I don't follow this person, I start following this person :D This feels weird as the action differs from the button-title/state.

Do you have an idea how to deal with this?

kimar:
@zeitschlag re the layout issue: I'm changing priority for the labels so the followers will take precedence over the verified link. Regarding the button state bug: I'm going to dig into this 👀

MastodonSDK/Sources/MastodonUI/View/Content/UserView.swift
DK1DA: This was done more or less intentionally as there was no need to expose the distinct type of button, but that said, as it's only used privately anyways it might be smarter to declare the actual type.

MastodonSDK/Sources/MastodonUI/View/Content/UserView+ViewModel.swift
DK1DA: Nope, no real reason, maybe I did some testing with this before actually using / implementing `MastodonMetricFormatter`, I'm happy to change this 👍

MastodonSDK/Sources/MastodonUI/Extension/NSAttributedString+Format.swift
DK1DA: No it didn't but it does now 😄

MastodonSDK/Sources/MastodonUI/View/Content/UserView.swift
DK1DA: Yes totally agree, I did an approach there but noticed that it's more work so I went with my current implementation.

MastodonSDK/Sources/MastodonUI/View/Content/UserView+ViewModel.swift
DK1DA: I agree that this should probably be more explicit so it won't be that easily forgotten. Happens to me as well.

Mastodon/Scene/Profile/FamiliarFollowers/FamiliarFollowersViewController.swift
DK1DA: It seems to be used in `DiscoveryForYouViewController` but I can't actually track this down to any actual usage, maybe it's some leftover code...

j-f1: It’s the last tab on the search & explore screen.

MastodonSDK/Sources/MastodonUI/View/Content/UserView.swift
zeitschlag: Alrighty :)

zeitschlag:
@kimar Unfortunately I can't start a review on this Github-PR, nor can I add single comments, I get `500`-status code all the time :/. So I'll just dump my thoughts and feedback here for now:

- We can't add a protocol extension to `UserTableViewCellDelegate` that implements that `DataSourceFacade.responseToUserViewButtonAction`-call due to the ViewController(s) being `self` as `dependency`-parameter, right?
- `MastodonAccountInMemoryCache` is the hacky workaround for the account-follow-issue, right? Looks like a very pragmatic approach 👍 
- The [`authorFollowersLabel`](https://github.com/mastodon/mastodon-ios/blob/93533654f8c08d4067ebe64570790ee0e1815d25/MastodonSDK/Sources/MastodonUI/View/Content/UserView%2BViewModel.swift#L93) doesn't support Dynamic Type (rest of the screen does):



- I consider this a very edgy edgecase: When I blocked an account and I search for that account, the 'Follow'-button is without an effect and it says 'Follow'. From my point of view it's not that big of a deal, I just wanted to mention it (as I read somethingsomething blocked accounts in the PR and I don't see blocked accounts in the following/followers-list)

zeitschlag:
@kimar Unfortunately I can't start a review on this Github-PR, nor can I add single comments, I get `500`-status code all the time :/. So I'll just dump my thoughts and feedback here for now:

- We can't add a protocol extension to `UserTableViewCellDelegate` that implements that `DataSourceFacade.responseToUserViewButtonAction`-call due to the ViewController(s) being `self` as `dependency`-parameter, right?
- `MastodonAccountInMemoryCache` is the hacky workaround for the account-follow-issue, right? Looks like a very pragmatic approach 👍 
- The [`authorFollowersLabel`](https://github.com/mastodon/mastodon-ios/blob/93533654f8c08d4067ebe64570790ee0e1815d25/MastodonSDK/Sources/MastodonUI/View/Content/UserView%2BViewModel.swift#L93) doesn't support Dynamic Type (rest of the screen does):



- I consider this a very edgy edgecase: When I blocked an account and I search for that account, the 'Follow'-button is without an effect and it says 'Follow'. From my point of view it's not that big of a deal, I just wanted to mention it (as I read somethingsomething blocked accounts in the PR and I don't see blocked accounts in the following/followers-list)

zeitschlag:
@kimar Unfortunately I can't start a review on this Github-PR, nor can I add single comments, I get `500`-status code all the time :/. So I'll just dump my thoughts and feedback here for now:

- We can't add a protocol extension to `UserTableViewCellDelegate` that implements that `DataSourceFacade.responseToUserViewButtonAction`-call due to the ViewController(s) being `self` as `dependency`-parameter, right?
- `MastodonAccountInMemoryCache` is the hacky workaround for the account-follow-issue, right? Looks like a very pragmatic approach 👍 
- The [`authorFollowersLabel`](https://github.com/mastodon/mastodon-ios/blob/93533654f8c08d4067ebe64570790ee0e1815d25/MastodonSDK/Sources/MastodonUI/View/Content/UserView%2BViewModel.swift#L93) doesn't support Dynamic Type (rest of the screen does):



- I consider this a very edgy edgecase: When I blocked an account and I search for that account, the 'Follow'-button is without an effect and it says 'Follow'. From my point of view it's not that big of a deal, I just wanted to mention it (as I read somethingsomething blocked accounts in the PR and I don't see blocked accounts in the following/followers-list)

zeitschlag:
@kimar Unfortunately I can't start a review on this Github-PR, nor can I add single comments, I get `500`-status code all the time :/. So I'll just dump my thoughts and feedback here for now:

- We can't add a protocol extension to `UserTableViewCellDelegate` that implements that `DataSourceFacade.responseToUserViewButtonAction`-call due to the ViewController(s) being `self` as `dependency`-parameter, right?
- `MastodonAccountInMemoryCache` is the hacky workaround for the account-follow-issue, right? Looks like a very pragmatic approach 👍 
- The [`authorFollowersLabel`](https://github.com/mastodon/mastodon-ios/blob/93533654f8c08d4067ebe64570790ee0e1815d25/MastodonSDK/Sources/MastodonUI/View/Content/UserView%2BViewModel.swift#L93) doesn't support Dynamic Type (rest of the screen does):



- I consider this a very edgy edgecase: When I blocked an account and I search for that account, the 'Follow'-button is without an effect and it says 'Follow'. From my point of view it's not that big of a deal, I just wanted to mention it (as I read somethingsomething blocked accounts in the PR and I don't see blocked accounts in the following/followers-list)

zeitschlag:
@kimar Unfortunately I can't start a review on this Github-PR, nor can I add single comments, I get `500`-status code all the time :/. So I'll just dump my thoughts and feedback here for now:

- We can't add a protocol extension to `UserTableViewCellDelegate` that implements that `DataSourceFacade.responseToUserViewButtonAction`-call due to the ViewController(s) being `self` as `dependency`-parameter, right?
- `MastodonAccountInMemoryCache` is the hacky workaround for the account-follow-issue, right? Looks like a very pragmatic approach 👍 
- The [`authorFollowersLabel`](https://github.com/mastodon/mastodon-ios/blob/93533654f8c08d4067ebe64570790ee0e1815d25/MastodonSDK/Sources/MastodonUI/View/Content/UserView%2BViewModel.swift#L93) doesn't support Dynamic Type (rest of the screen does):



- I consider this a very edgy edgecase: When I blocked an account and I search for that account, the 'Follow'-button is without an effect and it says 'Follow'. From my point of view it's not that big of a deal, I just wanted to mention it (as I read somethingsomething blocked accounts in the PR and I don't see blocked accounts in the following/followers-list)

Mastodon/Diffable/User/UserSection.swift
```diff
@@ -26,9 +27,10 @@ extension UserSection {
         weak var userTableViewCellDelegate: UserTableViewCellDelegate?
     }
 
-    static func diffableDataSource(
+    static  func diffableDataSource(
```
zeitschlag:
```suggestion
    static func diffableDataSource(
```

zeitschlag (approved):
Looks good, due to Github-status-related reasons I posted my notes here: 

https://github.com/mastodon/mastodon-ios/pull/1032#issuecomment-1540011935

kimar:
@zeitschlag thank you for the thorough review, regarding your points:

1. Very good catch! I've fixed this and made it part of the protocol extension, this indeed works.
2. Yes exactly, it's better this way than it was in my v1 but the ideal scenario would be if we could use the db for this. I had a look at how android does its and it seems they are also (only) keeping this in memory.
3. Thank you, this is fixed now.
4. I'm tasking a quick look at this.

kimar:
@zeitschlag again re #4:

The issue you've seen is due to the in-memory state being put of sync with the remote state, I've added an update to the state on (un)follow / block interactions, this should fix your issue: https://github.com/mastodon/mastodon-ios/pull/1032/commits/3e6d75e1b5f0bd7a03f27e1206750f85b87bef01

PR #1039, opened 2023-05-08
Use correct color when switch between darkmode/lightmode (#1034)

### LINKED ISSUE #1034: Textcolor in Hashtag-Widget doesn't change when toggling Darkmode/Lightmode
Black text isn't very readable on black background, and so is white text on a white background.

... something something HTML, I guess?

zeitschlag:
The issue was that when switching between darkmode and lightmode, the text color of the status-text didn't change. We used a HTML-template with some CSS to style hashtags, both was rendered into an `NSAttributedString` and then into an `AttributedString` and then into a SwiftUI-`Text`. Somewhere along that line, something didn't work as expected:

<video src="https://user-images.githubusercontent.com/2580019/236848592-2a10d98a-2963-47cc-b993-48d3f4fe29da.mp4"></video>

This is an alternative approach now, still rendering the status-HTML-content into `NSAttributedString`, but coloring the hashtags (using regex like a real #hacker). Now the color changes correctly when switching between lightmode and darkmode and vice versa:

<video src="https://user-images.githubusercontent.com/2580019/236848219-2e0034a1-0f5d-48f8-8eb9-a0a9d09c3149.mp4"></video>

Closes #1034


WidgetExtension/Variants/Hashtag/HashtagWidgetView.swift
```diff
@@ -81,54 +81,63 @@ struct HashtagWidgetView: View {
...
-                    color: \(accentColor);
-                }
+        let textColor = UIColor(named: "Colors/TextColor")
+        let accentColor = UIColor(named: "Colors/AccentColor")
+        let font = UIFontMetrics(forTextStyle: .headline).scaledFont(for: .systemFont(ofSize: fontSize, weight: fontWeight))
+
+        let attributedString: AttributedString
+
+        // 1. Render status-content as HTML ...
+        if let data = htmlString.data(using: .utf8),
```
DK1DA: 💡  Maybe use `guard let …` here to save on one indentation?

kimar (approved):
🚀

j-f1:
A couple of options here:

- If you use the `MastodonMetaContent` parser, you can use something like this to identify detected content (like links/hashtags/mentions) in the post and then update its properties:
  ```swift
  str.enumerateAttribute(NSAttributedString.Key("MetaAttributeKey.meta"), in: NSMakeRange(0, valueMetaLabel.textStorage.length)) { value, range, _ in
      if let value = value as? Meta {
          result.append((valueMetaLabel.textStorage.string.substring(with: range), value))
      }
  }
  ```
- Otherwise, the regex used should probably be identical to the one used in the web UI: https://github.com/mastodon/mastodon/blob/9818f342735d1765baa281aaeeab2f60b8d049fe/app/javascript/mastodon/utils/hashtags.ts

WidgetExtension/Variants/Hashtag/HashtagWidgetView.swift
zeitschlag: That's an idea, thank you :)

zeitschlag:
@j-f1 Uh, `MastodonMetaContent` sounds good, I'll have a look into that :) 

I didn't know that it exists, so thank you very much!

zeitschlag:
@j-f1 Using the `MetaContent`-stuff for highlighting was a great idea, thank you! I had to add some hacky hacks, like the [`attributedString()`-method](https://github.com/TwidereProject/MetaTextKit/blob/main/Sources/Meta/MetaContent.swift#L18) but in the end it worked and it feels much better this way!

(Emojis won't work yet, unfortunately.)

WidgetExtension/Variants/Hashtag/HashtagWidgetView.swift
```diff
@@ -89,55 +91,57 @@ struct HashtagWidgetView: View {
...
             // this is a fallback
-            attributedString = AttributedString(NSAttributedString(string: htmlString))
+            attributedString = AttributedString(NSAttributedString(string: statusContent))
         }
 
         self.init(attributedString)
     }
 }
+
+#warning("Replace this once we update MetaTextKit to `4.5.2`. We should do that.")
```
j-f1, line 123: MetaTextKit 3 and above use the new TextKit 2 APIs, and it seems like TextKit 2 has a bunch of bugs / missing functionality so @MainasuK didn’t recommend updating yet.

j-f1:
Yay, glad it worked! I think you’re intended to pass the raw HTML into MastodonMetaContent along with the emoji dictionary, so that may be why the emoji aren’t working.

WidgetExtension/Variants/Hashtag/HashtagWidgetView.swift
zeitschlag: Ah, I wasn't aware of that, again: Thank you :)

MainasuK: Yes. Although I fix some bugs like the [wrong RTL layout](https://github.com/TwidereProject/MetaTextKit/commit/c2a3d5e4355c1e3c9046c89433b28795f3374876) when layout using TextKit 2 in the latest MetaTextKit (v4). But the paragraph spacing still does not be respected by the text engine. 

I recommend using the native UITextView with NSAttributedString which export from the parser. Here is an [example](https://github.com/TwidereProject/TwidereX-iOS/blob/develop/TwidereSDK/Sources/TwidereUI/UIViewRepresentable/TextViewRepresentable.swift) I'm working on. And this requires iOS 16.0+. So staying in the **stable** old version and upgrading to this in the future is a good idea. :)

zeitschlag: Unfortunately we can't use UIKit-stuff in Widgets. It's just not possible :) So we'll just wait then? :D

MainasuK: @zeitschlag Oh. I think the custom emoji will not display if the UIKit is not usable in the Widget. Due to AttributedString doesn't support custom text attachments. Maybe it's acceptable.

kimar (approved):

kimar (approved):

PR #1045, opened 2023-05-13
Fix crash when a server has an emoji category named after its domain

### LINKED ISSUE #1042: App crashes when I want to reply
This has been as issue for me since the start of the year. I can’t send a reply with the app. 

Steps to reproduce. 

- Open app and logon
- Click message balloon on a toot 
- 💥

j-f1:
Fixes #1042 two ways:

- The uncategorized (named after the server) section can no longer conflict with a named category section in the emoji picker and cause a crash
- The uncategorized section is now only shown if there are uncategorized emoji on the server.

To test, create a c.im account or update `EmojiService+CustomEmojiViewModel+LoadState.swift:43` and `CustomEmojiPickerSection+Diffable.swift:49` to hardcode `"c.im"` as the server domain.


kimar (approved):
LGTM, really great work @j-f1 thanks for tracking this down and fixing it!

PR #1046, opened 2023-05-13
Remove prohibited calling the super method

woxtu:
Hi,

This PR removes prohibited calling `super.loadView()`.

The document says:

> Your custom implementation of this method should not call `super`.

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621454-loadview

### DIFF
--- a/Mastodon/Scene/MediaPreview/AltTextViewController.swift
+++ b/Mastodon/Scene/MediaPreview/AltTextViewController.swift
@@ -48,7 +48,6 @@ class AltTextViewController: UIViewController {
     }
     
     override func loadView() {
-        super.loadView()
         view.translatesAutoresizingMaskIntoConstraints = false
     }
 


j-f1:
Have you tested this out locally? I would assume that since we never assign to `view` this would cause an infinite loop when presenting the alt text popover.

woxtu:
Right. I should have read the following text:

> If you want to perform any additional initialization of your views, do so in the `viewDidLoad()` method.

I'll make a new PR.

j-f1:
I added the method in https://github.com/mastodon/mastodon-ios/commit/0faea45485693ae7f00c947f4f2100079ec8eb7c, I think the issue might have been that setting `translatesAutoresizingMaskIntoConstraints` didn’t work out if I did it in `viewDidLoad`, but if you can confirm that the constraints don’t get messed up when you move the line it should be ok to move it!

woxtu:
Thank you for your detailed reply!

PR #1047, opened 2023-05-13
Fix the alt text view initialization

woxtu:
Hi,

This PR fixes the alt text view initialization.

The document says:
> If you want to perform any additional initialization of your views, do so in the `viewDidLoad()` method.

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621454-loadview

Related to https://github.com/mastodon/mastodon-ios/pull/1046.

### DIFF
--- a/Mastodon/Scene/MediaPreview/AltTextViewController.swift
+++ b/Mastodon/Scene/MediaPreview/AltTextViewController.swift
@@ -47,15 +47,11 @@ class AltTextViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
     
-    override func loadView() {
-        super.loadView()
-        view.translatesAutoresizingMaskIntoConstraints = false
-    }
-
     override func viewDidLoad() {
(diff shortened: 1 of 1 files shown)


kimar:
@woxtu was there a specific issue this PR is resolving or is it just related to Apple's Developer Documentation? I tried it out and it seems to be fine but also it looks completely fine if `view.translatesAutoresizingMaskIntoConstraints = false`.
When looking at the implementation and usage of `AltTextViewController` I'm also wondering if this is needed at all.

@j-f1 do you remember if there was a specific reason you set `translatesAutoresizingMaskIntoConstraints` to `false` on the VC's view? Just wanting to double check to make sure we're safe.

j-f1:
I don’t remember why I did it this way and looking back at the commit where I added the `loadView` method yields no explanation either.

woxtu:
@kimar It's the latter.

kimar (approved):
LGTM 👍

PR #1048, opened 2023-05-15
Remember notifications tab after app restart

kimar:
# Rationale

Remember the most recently selected notifications tab after an app restart, depending on wether "Mentions" or "Everything" was selected.

### DIFF
--- a/Mastodon/Scene/Notification/NotificationViewModel.swift
+++ b/Mastodon/Scene/Notification/NotificationViewModel.swift
@@ -25,9 +25,26 @@ final class NotificationViewModel {
     // output
     let scopes = NotificationTimelineViewModel.Scope.allCases
     @Published var viewControllers: [UIViewController] = []
-    @Published var currentPageIndex = 0
-
+    @Published var currentPageIndex = 0 {
+        didSet {
+            lastPageIndex = currentPageIndex
+        }
+    }
     
+    private var lastPageIndex: Int {
+        get {
+            UserDefaults.shared.getLastSelectedNotificationsTab(
+                accessToken: authContext.mastodonAuthenticationBox.userAuthorization.accessToken
+            )
+        }
+        set {
+            UserDefaults.shared.setLastSelectedNotificationsTab(
+                accessToken: authContext.mastodonAuthenticationBox.userAuthorization.accessToken,
+                value: newValue
+            )
+        }
+    }
+
     init(context: AppContext, authContext: AuthContext) {
         self.context = context
         self.authContext = authContext
@@ -58,7 +75,14 @@ extension NotificationViewModel: PageboyViewControllerDataSource {
     }
     
     func defaultPage(for pageboyViewController: PageboyViewController) -> PageboyViewController.Page? {
-        return .first
+        guard
+            let pageCount = pageboyViewController.pageCount,
+            pageCount > 1,
+            (0...(pageCount - 1)).contains(lastPageIndex)
+        else {
+            return .first /// this should never happen, but in case we somehow manage to acquire invalid data in `lastPageIndex` let's make sure not to crash the app.
+        }
+        return .at(index: lastPageIndex)
     }
     
 }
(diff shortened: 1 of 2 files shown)


Mastodon/Scene/Notification/NotificationViewModel.swift
```diff
@@ -25,8 +26,14 @@ final class NotificationViewModel {
...
     @Published var viewControllers: [UIViewController] = []
-    @Published var currentPageIndex = 0
-
+    @Published var currentPageIndex = 0 {
+        didSet {
+            lastPageIndex = currentPageIndex
+        }
+    }
+    
+    @StoredUserDefaults(defaults: .shared, key: "lastSelectedNotificationsTabIndex", defaultValue: 0)
```
zeitschlag: When using multiple accounts, this changes the selected tab for all accounts visibly when you go to the notification-tab. Is this intended?

Mastodon/Scene/Notification/NotificationViewModel.swift
```diff
@@ -25,8 +26,14 @@ final class NotificationViewModel {
     // output
     let scopes = NotificationTimelineViewModel.Scope.allCases
     @Published var viewControllers: [UIViewController] = []
-    @Published var currentPageIndex = 0
-
+    @Published var currentPageIndex = 0 {
+        didSet {
+            lastPageIndex = currentPageIndex
```
zeitschlag, line 30: Using a property wrapper here instead of a `UserDefaults.shared.set(...)` or `UserDefaults.shared.integer` feels kind of overengineered as this is the only occasion where we use this (yet?).

Also this differs from the usual way the app uses `UserDefaults` (search for `extension UserDefaults`, for example :)

zeitschlag (commented):
Thank you @kimar! I do have an opinion here (see comments), but that doesn't mean, that you have to agree with me :) Looks good and does, what it should (mostly :D), so: Feel free to merge :)

Mastodon/Scene/Notification/NotificationViewModel.swift
DK1DA: Hm ,that's a very good point, I haven't taken there current usage of UserDefaults into account, I'll change it so it matches those.

Mastodon/Scene/Notification/NotificationViewModel.swift
DK1DA: Right, yet again something I didn't think about, however it does make sense to scope this on a per-account basis as this usage is arguably depending on your account and the number off notification it gets.

MastodonSDK/Sources/MastodonCommon/Preference/Preference+Notification.swift
```diff
@@ -38,6 +39,17 @@ extension UserDefaults {
         setNotificationCountWithAccessToken(accessToken: accessToken, value: count + 1)
     }
     
+    @objc public func getLastSelectedNotificationsTab(accessToken: String) -> Int {
+        let prefix = UserDefaults.notificationsLastTabIndexKeyPrefix
+        let key = UserDefaults.deriveKey(from: accessToken, prefix: prefix)
```
zeitschlag, line 44: Using `deriveKey` is clever here! I assume, the `accessToken` is permament in a sense that it doesn't change frequently? Just to make sure, that `lastSelectedNotificationsTab`-entries won't pile up in UserDefaults

zeitschlag (approved):
🚀

MastodonSDK/Sources/MastodonCommon/Preference/Preference+Notification.swift
DK1DA: @zeitschlag from what I've seen it's long lived and I haven't seen a refresh being made. Also this technique is used for other defaults as well (which probably isn't a perfect excuse 🥸)

zeitschlag: It's a good enough excude 🤓

PR #1051, opened 2023-05-23
Experiment: Switch CoreData to In-Memory Store for performance improvements

kimar:
# Playground

**No data will be persisted to CoreData and you'll need to re-authenticate after restarting the App.**

## Things, that need to be persisted otherwise:

- [x] Credentials ➡️ Keychain
- [ ] Settings
- [ ] Widgets can't use the Credentials right now


j-f1:
If you’re working on performance, check out https://github.com/mastodon/mastodon-ios/pull/817 — I found the cause of some hitches on thread screens.

kimar:
@j-f1 thank you! I'll definitely check it out.

zeitschlag:
As I'm working on IOS-14 (#1073) and this changes the settings, we might have to touch this part of the app (again). Logout might be affected, too.

PR #1053, opened 2023-05-29
Fix loading of follower/following count on profile pages

j-f1:
This PR reverts #825 and replaces it with an implementation that fetches the entire account object from the API. This allows us to ensure all information is up to date (including follow/follower counts, as well as name/bio/etc.


kimar (approved):
Very good! Thank you @j-f1 !

PR #1054, opened 2023-06-01
Fix Brand Color (IOS-160)

### LINKED ISSUE #938: Links hard to see in dark mode

### Current Behavior

The light purple used for links (username, web) has insufficient contrast with the light gray background when the app is set to dark mode (not “true black” mode). I suggest using a lighter color and verifying against WCAG standards. 

### Expected Behavior

Link text is easily readable. 

### Steps To Reproduce

1. Set app to dark mode (not “true black” mode)
2. View any post with one or more links

### Environment

```markdown
- Device:iPhone 14 Pro
- OS:16.3
- Version:1.4.11
- Build:265
```

### Anything else?

_No response_

zeitschlag:
We used the wrong accent color ("blurple") in Dark Mode (see the left two screenshots below). On the two screenshots on the right, you see the correct accent color being used for hashtags and the selected tab bar item:

ios_160

So, ideally, for Light Mode, nothing has changed, but Dark Mode looks better now :sparkl

While searching and replacing `brand.color` I noticed `ComposeStatusPollExpiresOptionCollectionViewCell.swift` not being used, so I removed it.

Closes #938


kimar (approved):
Very nice @zeitschlag 👏  🚀  Thanks!

PR #1055, opened 2023-06-02
Post Overflow Menus (IOS-103)

zeitschlag:
This is how it should look like:

Screenshot 2023-06-14 at 16 52 24

- Menus feature two sections now: Post actions and User actions. Each contains different items.
- Own posts have two sections: One for Post actions and one to delete a post.
- You can untranslate translated posts now
- You can follow/unfollow people now from that menu
- See some variants in this screenshot

ios-103-2

Please note, that the notification screen doesn't feature the new sections/items. This is mostly due to pragmatic reasons: Refactoring that screen would've taken way longer. I created an issue for that (IOS-170).

Also the "Add to list"-stuff is missing as we don't have lists... yet.

### DIFF
--- a/Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
+++ b/Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
@@ -187,6 +187,7 @@ extension NotificationView {
             }
             .assign(to: \.isBlocking, on: viewModel)
             .store(in: &disposeBag)
+
         // isMyself
         Publishers.CombineLatest(
             author.publisher(for: \.domain),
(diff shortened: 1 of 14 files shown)


zeitschlag:
Be aware, that there's a Github-actions after merges, that ships a new Testflight. As we (mis)use that to give the performance improvements a testdrive, we should:

1. Cancel the Build Testflight-job
2. Rebase the [`performance_experiments`](https://github.com/mastodon/mastodon-ios/tree/performance_experiments)-branch on `develop`.
3. Rebase the [`release-testflight-performance-improvements`](https://github.com/mastodon/mastodon-ios/tree/release-testflight-performance-improvements)-branch on `performance_experiments`

Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
```diff
@@ -199,12 +200,27 @@ extension NotificationView {
...
         notification.publisher(for: \.transientFollowRequestState)
             .assign(to: \.transientFollowRequestState, on: viewModel)
             .store(in: &disposeBag)
+
+        // Following
+        author.publisher(for: \.followingBy)
+            .map { [weak viewModel] followingBy in
+                guard let viewModel = viewModel else { return false }
+                guard let authContext = viewModel.authContext else { return false }
+                return followingBy.contains(where: {
```
DK1DA, line 218: ❓ I wonder if `followingBy` contains the correct data, I've had some issues with this in the past when implementing the following/followedBy state for the user lists.

kimar (approved):
LGTM 👍

Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
zeitschlag: Yeah, I get that. To be honest: I didn't know which to use either and so I tried both and this one worked 😬

PR #1019, opened 2023-04-16
IOS-148 Updates to the media badging look & feel

j-f1:
Individual badges

- [x] Badges are not fixed size; they are labels with padding and a background color
- [x] 8pt horizontal padding, 2pt vertical padding
- [x] Label font: `.font(.subheadline).bold()`
  - [x] Set label to use monospacedDigits [for duration only]
- [x] Label foreground color: always white
- [x] Badge background color: black, 0.7 alpha

Expanded badge state
- [x] Expanded width should take as much space as possible minus the 8pt insets on both horizontal edges
- [x] Badge expands with an animation on tap
- [x] Expands outwards on top of any other badges (if applicable) rather than pushing out other badges
- [x] Tapping anywhere ~~aside from the translation button~~ collapses the badge
- [x] 8pt padding, all edges
- [x] Label font: `.font(.caption1) `

Translation: [not implemented]

Group of badges

- [x] 2pt spacing between badges
- [x] Inset 16pt from leading edge of media container, 8pt from bottom
- [x] If there are multiple badges to display, the order (LTR) should be:
  - ALT
  - GIF
  - Video duration timestamp (MM:SS)

<details><summary>Screenshots</summary>

     

IMG_2924

</details>

PR #1058, opened 2023-06-04
IOS-119 Video attachment adjustments

j-f1:
- [x] This is not an inline player. Tapping the attachment opens the fullscreen media viewer.
  - Already the case
- [x] Play button is SF Symbol play.circle.fill. It is always white with a drop shadow.
  - Updated sizing, removed the backdrop!
- [x] Use a media badge (IOS-125) to display the video duration timestamp (MM:SS)
  - Implemented in #1019

Proposal: add a blur behind the triangle to improve contrast on complex videos (ultra thin material, always in the light color scheme regardless of app color scheme, on [this post](https://mastodon.social/@hootalex/110487091560971840) hacked to display the play button). What do you think @samhenrigold?

(included in this PR but willing to revert!)

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/InlineMediaOverlayContainer.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/InlineMediaOverlayContainer.swift
@@ -8,10 +8,16 @@ import SwiftUI
 
 struct InlineMediaOverlayContainer: View {
     var altDescription: String?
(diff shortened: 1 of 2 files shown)


Gargron:
I like the blur idea, fine from my side.

zeitschlag (approved):
Looks good from my side, too, thank you @j-f1 🎊

PR #1059, opened 2023-06-05
IOS-159: Fix tap location check for dismissing media preview

j-f1:
`view.bounds` is in the coordinate space of `view`, while `view.frame` is in the coordinate space of `view.superview`.


kimar (approved):
Great @j-f1 🚀  Thank you!

PR #1069, opened 2023-06-24
Set a solid background on the blurhash view in case it doesn’t load

### LINKED ISSUE #1067: [BUG] Blur delays on sensitive content

### Current Behavior

While scrolling through my feed in the app, while most sensitive content is properly blurred, some posts will not be. All non-blurred posts will still have the "tap to reveal" text overlay.
Seems like the blur filter is applied at a later step than the media itself, scrolling with any sort of speed leads to delays in the blur. The delay can be anywhere from just a second, to much longer, which could lead to some less than desirable outcomes.

This could be a copy of https://github.com/mastodon/mastodon-ios/issues/961
But it was unclear to me if all their content was not being properly handled, or if its the same delay issue I'm experiencing. The lack of comment on that post in two months made me hesitant.
More than happy to move this to there if that's preferred. 

### Expected Behavior

All sensitive content should be blurred prior to media being displayed, so no 'leaks' occur.

### Steps To Reproduce

1. In the app I searched for '#reddit'
2. Scroll with some speed due to all the sensitive content
3. Some media should (at least on my end) load before the blur has been applied.

### Environment

```markdown
- Device:Iphone 12
- OS:IOS 16.1.1
- Version:v2023.9
- Build:324
```

### Anything else?

Not gonna post nsfw content here, but I could try a screen capture and send that through another channel if it would help.

j-f1:
Fixes #1067. It seems that the underlying issue there was that the NSFW media somehow loaded before the blurhash image. (Maybe because whatever process produces the blurhashes gets backed up when scrolling quickly?) I fixed this by setting the background color of the blurhash image to a solid gray. This prevents anything from leaking through.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/MediaView.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/MediaView.swift
@@ -30,6 +30,8 @@ public final class MediaView: UIView {
         imageView.contentMode = .scaleAspectFill
         imageView.isUserInteractionEnabled = false
         imageView.layer.masksToBounds = true    // clip overflow
+        imageView.backgroundColor = .gray
+        imageView.isOpaque = true
         return imageView
     }()
     


kimar (approved):
Great idea for a quick fix @j-f1 👍

PR #1070, opened 2023-06-24
Hide the entire top toolbar for videos & audio

### LINKED ISSUE #485: [BUG] Close Fullscreen Video

### Current Behavior

I encountered, when watching a video on mastodon iOS (iphone X) in fullscreen there is no possibility to close the window. So you can only close the whole app and restart.

### Expected Behavior

A close button would be nice.
Keep up the good work!

### Steps To Reproduce

1. open any video in the app
2. try to close it 😁
3. …
4. (decide you watch too many videos, anyways)
5. (have an epiphany on how to prevent the climate catastrophe)
6. (quit doomscrolling and use your time for the greater good)

### Environment

```markdown
- Device: iphone X
- OS: iOS
- Version: 1.4.5
- Build:
```

### Anything else?

_No response_

### LINKED ISSUE #955: [BUG] Cannot Exit Videos

### Current Behavior

When opening a video in the Mastodon App, I have found myself unable to exit the fullscreen mode by clicking on the “x” button. The only way to exit is to close the entire app. 

### Expected Behavior

Hitting the button should close the video. 

### Steps To Reproduce

1. Open a video
2. Tap the close button 
3. Tap the close button 
4. …

### Environment

```markdown
- Device: Apple iPhone 13
- OS: 16.2
- Version: v1.4.11
- Build: 265
```

### Anything else?

_No response_

j-f1:
Hopefully fixes #485, fixes #955

Currently, the “alt” button covers the entire top of the video player interface. This is bad because the close button is underlapped by the “alt” button. It would be awesome if there was a way to add a “media description” button or similar to the “…” menu in the video player but unfortunately those sorts of APIs are restricted to tvOS. Since the alt text can generally be read in the inline preview (especially since videos are less commonly present with other media attachments so there should be a bunch of space), and folks who use screen readers will already have access to the full text through their screen reader / other accessibility tool, I don’t think losing access to the popover UI will be a huge loss. I think that if y’all want to explore adding any sort of UI over top of the video player, it would be best to identify and switch to a high-quality, accessible third party video player that offers better customization options (if such a project exists).

### DIFF
--- a/Mastodon/Scene/MediaPreview/MediaPreviewViewController.swift
+++ b/Mastodon/Scene/MediaPreview/MediaPreviewViewController.swift
@@ -116,7 +116,7 @@ extension MediaPreviewViewController {
                     self.topToolbar.isHidden = {
                         guard index < previewContext.attachments.count else { return false }
                         let attachment = previewContext.attachments[index]
-                        return attachment.kind == .video    // not hide buttno for audio
+                        return attachment.kind == .video || attachment.kind == .audio
                     }()
                 default:
                     break
--- a/Mastodon/Scene/MediaPreview/MediaPreviewViewController.swift
+++ b/Mastodon/Scene/MediaPreview/MediaPreviewViewController.swift
@@ -113,12 +113,11 @@ extension MediaPreviewViewController {
                 guard let self = self else { return }
                 switch self.viewModel.item {
                 case .attachment(let previewContext):
-                    let needsHideCloseButton: Bool = {
+                    self.topToolbar.isHidden = {
                         guard index < previewContext.attachments.count else { return false }
                         let attachment = previewContext.attachments[index]
                         return attachment.kind == .video    // not hide buttno for audio
                     }()
-                    self.topToolbar.isHidden = needsHideCloseButton
                 default:
                     break
                 }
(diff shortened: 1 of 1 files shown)


Mastodon/Scene/MediaPreview/MediaPreviewViewController.swift
```diff
@@ -118,7 +118,7 @@ extension MediaPreviewViewController {
                         let attachment = previewContext.attachments[index]
                         return attachment.kind == .video    // not hide buttno for audio
                     }()
-                    self.closeButton.isHidden = needsHideCloseButton
+                    self.topToolbar.isHidden = needsHideCloseButton
```
zeitschlag: You could rename `needsHideCloseButton` to `topToolbarIsHidden` or something similar as it hides the toolbar now and not just the close button.

zeitschlag (approved):
@j-f1 Thank you for this quick fix

I'm a bit uncertain. On one hand, hiding the alt-button for video isn't most probably a big loss, as you said, but on the other hand, it's still something where video differs from a GIF or image. Also some people, well, at least myself, tbh, use the alt-text to add additional information, so having this alt-button there is pretty convenient.

But: I wouldn't add another external dependency just to solve this, we could wait two years or so, maybe Apple opens the API?

Either way: LGTM! 🚀

Mastodon/Scene/MediaPreview/MediaPreviewViewController.swift
j-f1: Updated to remove the variable entirely & also hid the top toolbar for audio since it uses the video player too.

zeitschlag: Great, that you saw and fixed this for audio, too 🥳 👍

PR #1073, opened 2023-06-27
Migrate existing settings to new Design

zeitschlag:
This is the first part for IOS-14. It migrates the existing settings over to new design. And it looks like this:


Mastodon/Scene/Settings/Notification Settings/Cells/NotificationSettingsDisabledTableViewCell.swift
```diff
@@ -0,0 +1,67 @@
...
+        backgroundColor = Asset.Colors.Brand.blurple.color.withAlphaComponent(0.15)
+
+        setupConstraints()
+    }
+
+    required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
+
+    private func setupConstraints() {
+        let constraints: [NSLayoutConstraint] = [
+            appBadgeImageView.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 16),
```
DK1DA: I feel like we should offload the `8` and `16` into some constants enum etc 😅

DK1DA: But we can do this later, no need to touch all those places again right now.

Podfile
```diff
@@ -17,8 +17,6 @@ target 'Mastodon' do
   pod 'Kanna', '~> 5.2.2'
   pod 'Sourcery', '~> 1.9'
 
-  # DEBUG
```
DK1DA, line 20: 👍

MastodonSDK/Sources/CoreDataStack/Entity/App/Setting.swift
```diff
@@ -14,6 +14,7 @@ public final class Setting: NSManagedObject {
     @NSManaged public var userID: String
 
 //    @NSManaged public var appearanceRaw: String
```
DK1DA: Shouldn't we get rid of this as well?

kimar (approved):
LGTM 👍  Very nice

MastodonSDK/Sources/CoreDataStack/Entity/App/Setting.swift
zeitschlag: Oh, yeah, you're right. Good point! 👍

PR #1077, opened 2023-06-30
Retry server search without the category if there are no results

j-f1:
From the veganism.social admin in the Discord:

> Can we remove the category filter when searching for a server by text on the mobile app? It's not obvious to new users that a filter is being applied.

If there are no results, the server search will be retried with the category filter removed:

(second screenshot is with the language filter set to Catalan)


numemac:
Awesome! Thank you! <3

zeitschlag:
@samhenrigold What do you think?

PR #1078, opened 2023-06-30
Improve heuristic for skipping the post discard confirmation modal

### LINKED ISSUE #1076: [BUG] When cancelling a new comment or reply, don't show the Discard option unless some text input is present.

### Current Behavior

If a user accidentally presses the comment or reply button and tries to cancel the comment without having typed anything, they will be presented with a "Discard?" prompt.

### Expected Behavior

If a user accidentally presses the comment or reply button and tries to cancel the comment without having typed anything, they will NOT be presented with a "Discard?" prompt since there is nothing to discard. It should only show the discard option if something has been typed (Apple Mail exhibits this behavior nicely)

### Steps To Reproduce

_No response_

### Environment

```markdown
- Device: iPhone 12 mini
- OS:
- Version:
- Build:
```

### Anything else?

_No response_

j-f1:
Fix #1076

Also expands the logic to show the confirmation modal if the user has entered a content warning, toggled the poll on, or added one or more attachments.

Additionally, removes some discard-related strings that are no longer used.

### DIFF
--- a/Localization/StringsConvertor/input/Base.lproj/app.json
+++ b/Localization/StringsConvertor/input/Base.lproj/app.json
@@ -15,10 +15,6 @@
                 "title": "Vote Failure",
                 "poll_ended": "The poll has ended"
             },
-            "discard_post_content": {
-                "title": "Discard Draft",
-                "message": "Confirm to discard composed post content."
-            },
             "publish_post_failure": {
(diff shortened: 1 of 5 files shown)


zeitschlag (approved):
LGTM, thank you @j-f1, also for cleaning up! 🚀 🥳

PR #1088, opened 2023-07-04
Hide activity indicator in case of no emojis

### LINKED ISSUE #1087: [BUG] custom emojis panel keeps loading

### Current Behavior

Users of our instance **berlin.social** reported about having issues to use the custom emojis panel of compose view. It keeps loading. It's reproducible for us when using this application. Strangely, it is reproducible using "Toot!" app, either. In browser, there are emojis, but these are just Unicode emojis, aren't they?

### Expected Behavior

Either a list of custom emojis appears or the loading indicator vanishes.

### Steps To Reproduce

1. Open app
2. Start composing a toot
3. Tap on smiley icon to open panel of custom emojis.

### Environment

```markdown
- Device: iPhone 13 Pro
- OS: 16.5.1
- Version: v2023.11
- Build: 337
```

### Anything else?

Tried accessing endpoint `GET /api/v1/custom_emojis` using authenticated browser as well as non-authenticated browser. In either case, empty list `[]` was returned, though request was pending for a few seconds when using authenticated browser. 

The instance has no local custom emojis, but there is a list of 1750 pages each listing a bunch of federation emojis in related view of web app's admin section. I assume this might cause the delay experienced when querying endpoint with an authenticated browser, but I doubt it's fetching that list thus taking the time.

Tried re-precompiling assets as found in a related but closed issue from 2021 to no avail.

Tried monitoring the log of mastodon-web service but there haven't been any log entries on queries sent to that endpoint. Also tried watching it for records with `status=` and a code that isn't starting with a `2`. But this didn't yield any access issues, either.

zeitschlag:
Closes #1087.


zeitschlag:
@j-f1 Hey, may I ask you for a review on this one? I'd be very thankful!

j-f1 (approved):
LGTM! It might be nice to add something saying “No emoji” but I think a blank screen gets the point across & is much better than having a loading indicator.

PR #1089, opened 2023-07-04
Restore zoom behavior when double-tapping on text in an image

### LINKED ISSUE #1084: [Feature]: double tap should enlarge the photo, not copy the text
### Description

I often enlarge photos. Double tap should enlarge the photo, not copy the text

https://github.com/mastodon/mastodon-ios/assets/53906414/6de836e1-f660-4ce5-b5d4-c9b6f9240b80

### Suggested Solution

Double tap to enlarge a photo, long tap to copy text

https://github.com/mastodon/mastodon-ios/assets/53906414/3d36f652-db1a-4839-8fc7-bfaac8334392

### Alternatives

Double tap to enlarge the photo, click on the icon in the lower right corner to copy the text. I think it was the solution of the Apollo app (for Reddit)

https://github.com/mastodon/mastodon-ios/assets/53906414/02e0de26-01ff-455e-81cd-de60cf0eb8ca

### Additional Context

_No response_

j-f1:
I added a delegate to the double-tap gesture recognizer that causes it to cancel any other double-tap gesture recognizers, except when the Live Text button is toggled on. This makes it easy to zoom in/out by double-tapping anywhere, while still allowing the double-tap-to-select gesture when the user has explicitly opted into highlighting the text in the image.

This may stop working at some point if the `ImageAnalysisInteraction` stops using a `UITapGestureRecognizer` internally, but I don’t think this is a big problem since it will just go back to the old behavior and not, like, crash or something. 

I also took the opportunity to rip out a long press gesture recognizer that was added to the `MediaPreviewImageView` but didn’t do anything.

Fixes #1084.

### DIFF
--- a/Mastodon/Scene/MediaPreview/Image/MediaPreviewImageView.swift
+++ b/Mastodon/Scene/MediaPreview/Image/MediaPreviewImageView.swift
@@ -70,6 +70,8 @@ extension MediaPreviewImageView {
         addSubview(imageView)
 
         doubleTapGestureRecognizer.addTarget(self, action: #selector(MediaPreviewImageView.doubleTapGestureRecognizerHandler(_:)))
+        doubleTapGestureRecognizer.delegate = self
+
         imageView.addGestureRecognizer(doubleTapGestureRecognizer)
         if #available(iOS 16.0, *) {
             imageView.addInteraction(liveTextInteraction)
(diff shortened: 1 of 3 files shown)


zeitschlag (approved):
Nice! Thank you @j-f1

PR #1115, opened 2023-09-15
Better Search Workflow (IOS-141)

zeitschlag:
We improved the Search Flow and were inspired by [the amazing Ivory](https://tapbots.com/ivory/).

Search History looks like this now:

It features a condensed version of the `UserView`.

When you search for things, there are two groups now. The upper one offers some handy shortcuts, while the lower group suggests up to three hashtags and up to three users:

If you paste a link into the hashtag, another elements becomes visible:

When this links leads to a status or user, you can see them directly inside the app. Oh, and we added some error-handling, too. Let's say, the user you're looking for doesn't exist, then alert tells you so:

But wait, that's not all! You can search for users on other servers, too. Just enter their handle...

... and the "Go to..."-option shows the way to them right away :)


Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -127,227 +127,6 @@ extension DataSourceFacade {
         let barButtonItem: UIBarButtonItem?
     }
     
-//    @MainActor
```
DK1DA, line 130: 🙌

Mastodon/Protocol/Provider/DataSourceFacade+SearchHistory.swift
```diff
@@ -8,11 +8,12 @@
 import Foundation
 import CoreDataStack
 import MastodonCore
+import UIKit
 
 extension DataSourceFacade {
     
     static func responseToCreateSearchHistory(
-        provider: DataSourceProvider & AuthContextProvider,
+        provider: NeedsDependency & UIViewController & AuthContextProvider,
```
DK1DA: We're using `NeedsDependency & UIViewController` quite often, maybe it makes sense to create a `typealias` for this?

Mastodon/Scene/Search/SearchDetail/SearchDetailViewController.swift
```diff
@@ -23,22 +22,20 @@ final class CustomSearchController: UISearchController {
 
 // Fake search bar not works on iPad with UISplitViewController
 // check device and fallback to standard UISearchController
-final class SearchDetailViewController: PageboyViewController, NeedsDependency {
```
DK1DA, line 26: Yes, I like that we got rid of `Pageboy`here. 👏

kimar (approved):
Looks really great @zeitschlag just one tiny thing which, when using "Posts matching XYZ" (or People matching XYZ) there is a certain delay between showing the view and when the activity indicator appears. In the first place to me this looked like there were no results and I just saw an empty screen. Maybe we can show the activity indicator immediate once the screen is show but has no results yet?

Mastodon/Protocol/Provider/DataSourceFacade+SearchHistory.swift
zeitschlag: it's.... 🥁 🥁 🥁 

🎉 `ViewControllerWithDependencies` 🎉

PR #1120, opened 2023-09-21
Clean dependencies and logging

zeitschlag:
- remove unused dependencies
- reduce logging


NotificationService/NotificationService.swift
```diff
@@ -75,10 +70,9 @@ class NotificationService: UNNotificationServiceExtension {
                 ImageDownloader.default.download(URLRequest(url: url), completion: { [weak self] response in
                     guard let _ = self else { return }
                     switch response.result {
-                    case .failure(let error):
-                        os_log(.info, log: .debug, "%{public}s[%{public}ld], %{public}s: download image %s fail: %s", ((#file as NSString).lastPathComponent), #line, #function, url.debugDescription, error.localizedDescription)
```
DK1DA, line 79: I'm wondering if this might actually be worth keeping...but I'm not sure how valuable it is given logs from the `NotificationService` are most likely rarely looked at anyways.

kimar (approved):
LGTM 🚀  So many changes 🤯  But it's good we're doing this!

NotificationService/NotificationService.swift
zeitschlag: Exactly, that's what I thought in most places: We don't do anything with these information. Logging an Error to Debug isn't very helpful. That's why I removed this (in several places)

PR #1121, opened 2023-09-22
Simplify Theme

zeitschlag:
We don't support custom themes anymore.


ShareActionExtension/Scene/ShareViewController.swift
```diff
@@ -69,16 +69,9 @@ extension ShareViewController {
...
-        ThemeService.shared.currentTheme
-            .receive(on: DispatchQueue.main)
-            .sink { [weak self] theme in
-                guard let self = self else { return }
-                self.setupTheme(theme: theme)
-            }
-            .store(in: &disposeBag)
-        
+        setupTheme()
+        ThemeService.shared.apply()
```
DK1DA, line 73: Do you know why we have to call `ThemeService.shared.apply()` here? This method is initially invoked in the `AppDelegate`.

kimar (approved):
LGTM 👍  As said, I'm a fan of removing this.

ShareActionExtension/Scene/ShareViewController.swift
zeitschlag: Probably cause it's an extension and I don't know if the AppDelegate gets invoked then?

DK1DA: Ah right, I didn't notice that it's part of the Share Extension. Makes sense then.

PR #1122, opened 2023-09-22
Update Gems to make use of CocoaPods 1.12.1

kimar:
# Rationale

The current `Gemfile.lock` has already been generated using CocoaPods v1.12.1 whereas our Gems still point to an older version. Let's fix this.


zeitschlag (approved):
Thank you for this fix!

PR #1124, opened 2023-09-25
Disable "Follow all"-button if there are no suggestions

zeitschlag:
If you don't follow any people, you can see some suggested accounts with a convenient "Follow all" at the bottom. This button was enabled all the time and while loading although this doesn't make a lot of sense. It looks like this now:

I noticed this while testing, so I decided to quickly fix it. And while doing so I decided to replace the `FollowButton` with a `UIButton` with a `UIButton.Configuration`, this fixed one warning. We could use this as a first step to migrate the `FollowButton` to make it work with `Configuration`.


kimar (approved):
🚀  LGTM!

PR #1126, opened 2023-09-28
Server Details in Settings (IOS-20)

zeitschlag:
We want to provide the user more information about their instance. That's why in the new settings, there's a section about exactly this. It's called ✨Server Details✨

ios_20

We show the rules and the extended description so that they don't have to go to the website to see the responsible people. From this screen they also can send an email directly to their adminstrator(s).

## Side Effects

- Removed unused packages and images
- Migrated [XLPagerTabStrip](https://github.com/xmartlabs/XLPagerTabStrip) from Cocoapods over to SPM


Mastodon/Scene/Onboarding/ServerRules/Cell/ServerRulesTableViewCell.swift
```diff
@@ -10,7 +10,7 @@ import MastodonAsset
 import MastodonLocalization
 
 final class ServerRulesTableViewCell: UITableViewCell {
-    
+    static let reuseIdentifier = "ServerRulesTableViewCell"
```
DK1DA, line 13: I'm a bigger fan of
```suggestion
    static let reuseIdentifier = String(describing: ServerRulesTableViewCell.self)
```

but feel free to leave as-is.

Mastodon/Scene/Settings/About Mastodon/AboutSettings.swift
```diff
@@ -23,7 +23,7 @@ enum AboutSettingsEntry: Hashable {
         case .privacyPolicy:
             return L10n.Scene.Settings.AboutMastodon.privacyPolicy
         case .clearMediaCache(_):
-            return L10n.Scene.Settings.AboutMastodon.cleaerMediaStorage
+            return L10n.Scene.Settings.AboutMastodon.clearMediaStorage
```
DK1DA, line 26: 💯

Mastodon/Scene/Settings/Server Details/AboutInstanceViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        tableView.tableHeaderView = headerView
+        tableView.tableFooterView = footerView
+
+        headerView.frame.size.height = 1
+        footerView.frame.size.height = 2
```
DK1DA: Can we make those heights part of `AboutInstanceTableHeaderView` and `AboutInstanceTableFooterView` construction process rather than leaking this here? Or are they re-used and this height altered dynamically? In this case please disregard.

kimar (approved):
Great work @zeitschlag LGTM 🚀

PR #1130, opened 2023-10-05
Implement Fastlane

kimar:
# Rationale

This PR adds Fastlane and is handling Codesigning and Provisioningprofile management via Match.

# On build numbers

With our previous build scripts we sometimes ran into race conditions where build numbers where assigned non-deterministically due to asynchroneously fetching them from existing builds on App Store Connect. This will be resolved as builds numbers are now determined by the git commit count.

# Sideeffects

* Upgrades CocoaPods to v1.13.0
* Upgrades various Gems


.github/workflows/develop-build.yml.inactive
zeitschlag, line 1: If we don't need it anymore, could we remove it then?

Documentation/Deployment.md
```diff
@@ -0,0 +1,3 @@
+# Deployment
+
+We're using [Fastlane](https://fastlane.tools) to deploy the App to App Store Connect. Please see the [Fastlane README](../fastlane/README.md) on the available commands.
```
zeitschlag, line 21: Maybe we (okay, you) could add a little paragraph what needs to be done for new developers (register their devices in `fastlane/devices.txt`...)

zeitschlag (approved):
LGTM!

Feel free to ignore the suggestions :)

.github/workflows/develop-build.yml.inactive
DK1DA: I've created a new Workflow to replace the old CI one and remove the legacy files.

Documentation/Deployment.md
DK1DA: Yes, this file specifically is only meant for App Store deployments, but I'll add a bit more info on local deployment as well.

fastlane/devices.txt
```diff
@@ -0,0 +1,3 @@
+Device ID	Device Name	Device Platform
```
zeitschlag: What came to my mind just yesterday: Is there a way to _not_ expose our devices publicly and still use fastlane?

DK1DA: We could add the `devices.txt` to the `.gitignore` afaik the devices will not be un-registered if you remove them anyways, so you could add your own devices to the file. use `bundle exec fastlane ios register_devices` but not push the file. This would add your device to the developer portal but not touch any existing devices.

PR #1131, opened 2023-10-11
Use Keychain for credentials

zeitschlag:
- Basically the [`performance_experiments`](https://github.com/mastodon/mastodon-ios/pull/1051)-branch, but with CoreData on disk instead of in memory and with the most recent `develop` merged.
- Migration: Move the credentials from CoreData to Keychain (and store in `UserDefaults` if migration was successful)
- Store new credentials in Keychain right away

## Side Effects

- Kill some MainThread-warnings


Mastodon/Scene/Profile/MeProfileViewModel.swift
```diff
@@ -29,5 +29,27 @@ final class MeProfileViewModel: ProfileViewModel {
...
+                try await context.managedObjectContext.performChanges {
+                    guard let me = self.authContext.mastodonAuthenticationBox.authentication.user(in: self.context.managedObjectContext) else {
+                        assertionFailure()
+                        return
+                    }
+
+                    self.me = me
+                }
+            } catch {
+                // do nothing?
```
DK1DA, line 51: Hm, maybe we should throw some error at the user, but not sure what the reason for such errors might be and if they're resolvable. Maybe let's create an issue for this to resolve this later?

kimar (approved):
👍 This is really nice and works as expected for me, tested the following scenarios:

1. Installed current App Store with Logged in Accounts
2. Installed this PR build -> Migrated as expected
3. Uninstalled App
4. Installed App Store build -> Asked to sign in
5. Uninstalled App Store build
6. Installed this PR build -> Instant sign in due to credentials being persisted in Keychain

LGTM to me 🚀

PR #1135, opened 2023-10-16
Don't persist translations

zeitschlag:
aka. Don't use CoreData for translations.

We don't need to persist translations. They're temporary nevertheless.

-------

This is basically my starting point for the CoreData-removal. Instead of ripping out something big, I decided to replace one property. I didn't feel confident to remove this property on a branch that doesn't build, that's why I chose `develop`. Being able to build and run to see if I may have broken the functionality helped me a lot.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/StatusAuthorView.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/StatusAuthorView.swift
@@ -179,7 +179,10 @@ extension StatusAuthorView {
             postActions.append(.editStatus)
         }
 
-        if let statusLanguage = menuContext.statusLanguage, menuContext.isTranslationEnabled {
+        if menuContext.isTranslationEnabled,
+           let statusLanguage = menuContext.statusLanguage,
+           let deviceLanguage = Bundle.main.preferredLocalizations.first,
+           deviceLanguage != statusLanguage {
             if menuContext.isTranslated == false {
                 postActions.append(.translateStatus(.init(language: statusLanguage)))
             } else {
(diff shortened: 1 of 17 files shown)


zeitschlag:
I want to add two more things:

- send device language identifier as `lang` to `translate`-request
- Hide the translation option, if the source-language of a status is equal to the device language.

Thank you @grishka!

kimar (approved):
Great work @zeitschlag 👏  LGTM

zeitschlag:
@kimar Thank you very much! Even though you already approved this pr, may I ask you for another short review of the last two commits?

MastodonSDK/Sources/MastodonUI/View/Content/StatusAuthorView.swift
```diff
@@ -179,7 +179,10 @@ extension StatusAuthorView {
             postActions.append(.editStatus)
         }
 
-        if let statusLanguage = menuContext.statusLanguage, menuContext.isTranslationEnabled {
+        if menuContext.isTranslationEnabled,
```
DK1DA, line 182: This doesn't look like a fault from this PR, but for some reason, on my self-hosted instance and als mastodon.social for me `menuContext.isTranslationEnabled` is always `false` thus I'm not seeing the menu entry 😕  But if it worked for you then I'm happy to move on. I'll investigate what's wrong on my side then...

kimar (approved):
PR LGTM but I had a general translation issue. Maybe you can make sure this is something that doesn't affect you.

MastodonSDK/Sources/MastodonUI/View/Content/StatusAuthorView.swift
zeitschlag: Thank you for the feedback, I'll have a closer look at it as it happened to me now, too. oO

zeitschlag: After removing the CoreData-Queue-stuff, it worked for me(tm). Can you give it another testrun, whenever you have a few minutes? That would be great.

Oh, sorry, I meant:

830jp7

kimar:
@zeitschlag Works for me now 👍

PR #1136, opened 2023-10-19
Don't persist following

zeitschlag:
tl;dr: Don't use `MastodonUser` on "Following"-screen, but `Mastodon.Entity.Account`

This PR does several things:

- Add [MBProgressHUD](https://github.com/jdg/MBProgressHUD) (cause I'm lazy) to show loading-states
- It fetches [the accounts that an account follows](https://docs.joinmastodon.org/methods/accounts/#followers) and the [relationships](https://docs.joinmastodon.org/methods/accounts/#relationships)
- It improves the `Loading`-state of the `FollowButton` in `UserView`
- Speaking of which: The `FollowButton` uses configuration now (which makes showing an ActivityIndicator very easy). Oh, the `FollowButton` is gone and got replaced by a basic `UIButton`.
- It migrates Kanna from Pods to SPM (We could think about removing Cocoapods, we now only use it for tooling aka SwiftGen and Sourcery and those are available via homebrew etc., too)


Mastodon/Scene/Profile/Following/FollowingListViewModel+State.swift
```diff
@@ -117,58 +124,72 @@ extension FollowingListViewModel.State {
...
                         maxID: maxID,
                         authenticationBox: viewModel.authContext.mastodonAuthenticationBox
                     )
 
                     var hasNewAppend = false
-                    var userIDs = viewModel.userFetchedResultsController.userIDs
-                    for user in response.value {
-                        guard !userIDs.contains(user.id) else { continue }
-                        userIDs.append(user.id)
+
```
DK1DA, line 151: I've run into an issue where there was a recursion in this state machine, leading to repeated calls to the relationship and followers APIs, in my case the followers list was empty so a call to the relationships was redundant, you might want to consider something like the following:

```suggestion
                    if accountResponse.value.isEmpty {
                        viewModel.accounts = []
                        viewModel.relationships = []
                        return await enter(state: Idle.self)
                    }
                    
```

kimar (changes requested):
When trying to view the followers of any account in my timeline, I'm only seeing an infinite spinner but no results. Looking at my own followers (Account Tab) works.

This might be related to the fact that, we're not showing following people which are on another server than the currently logged in instance. (In the current prod version I'm seeing the message in the screenshot instead).



Also I'm seeing lots of JSON decoding errors when going back to my Timeline (Mastodon Server v4.2.1):

Simulator Screenshot - iPhone 12 mini - 2023-10-30 at 14 53 17


```
Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "id", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"id\", intValue: nil) (\"id\").", underlyingError: nil))
```

and

```
Swift.DecodingError.typeMismatch(Swift.Array<Any>, Swift.DecodingError.Context(codingPath: [], debugDescription: "Expected to decode Array<Any> but found a dictionary instead.", underlyingError: nil))
```

kimar:
I've used **MBProgressHUD** in the past as well but I'm not a huge fan of pulling in a dependency which hasn't been updated in years, but also I totally feel the pain of not wanting to write an indicator from scratch.

@zeitschlag what do you think of creating a light-weight wrapper around **UIAlertController** to server as a modal loading indicator?

Mastodon/Scene/Profile/Following/FollowingListViewModel+State.swift
zeitschlag: Ah, great, I'll have a look at it, thank you!

zeitschlag: Unfortunately, this seems to be only half of the fix, the other half seems to be running the `ListBatchFetchViewModel` running amok — for whatever reason. I'll need to wrap my head around. Good, that you fetched this annoying bug!!

zeitschlag:
@kimar Thank you for your review :) Regarding `MBProgressHUD`: I'm not a huge fan of pulling in dependencies (regardless of their age) either, but in this case, I'd consider it a pragmatic way:

- It does basic stuff, nothing critical. So in case, it might stop working (to me, `MBProgressHUD` has proven to be a reliable library for years), we know how to easily replace it (create a light-weith wrapper that does the same thing and handle presentation).
- It hasn't been updated in three years, yes, but I'd consider it more due to a "It does its thing good enough" than "Noone wants to take care of it"
- I wanted to focus on the task at hand (migrating the following-list to entities and building endless loops :D) than adding a new, fancy feature.

We can still replace it with something better/nicer/... at any time, I'd have nothing against it and I'd encourage you (or everybody else reading this) to suggest/build something better/nicer/... :)

Mastodon/Scene/Profile/Following/FollowingListViewModel+State.swift
zeitschlag: Regarding the recursion: Great suggestion with entering the `Idle`-state, I picked `NoMore` so that it shows the "Follows from another servers"-text again.

And, as I said: I wondered why those requests keep getting sent every second or so. Turns out: There's a timer in `ListBatchFetchViewModel` which fires every second, that's why the activity indicator kept spinning. To me, this looks like a hack: Every second, this timer checks, if a refresh is needed.

I'll replace it for this very screen with another approach (did someone say `UIScrollViewDelegate`?), but using this timer-based approach seems to be quite a hack to me. We should think about a viable solution for the other screens.

DK1DA: Good findings! Sounds like we're getting somewhere. And yes I think we should move to a more reactive approach instead of a polling timer.

zeitschlag:
Regarding the `DecodingErrors`: It's probably connected to 02207d1b. I'll have a look at them. Also I'll also check the slighty higher memory usage. I didn't notice that before, could be connected to the images?

The reason for the `DecodingErrors` were issues with domains. On my machine, it works now™️

kimar (approved):
Tested and LGTM 🚀  Looks solid to me now.

PR #1137, opened 2023-10-25
IOS-174: Don't persist followers

zeitschlag:
tl;dr: Don't use `MastodonUser` on "Followers"-screen, but `Mastodon.Entity.Account`

- #1136, but for followers (not following)
- ~~This PR build on #1136, but should be merged in `develop`, once #1136 is merged~~, this was done automatically 🤯


kimar (approved):
LGTM 👍  I was just wondering if we could even get rid of `MastodonAuthenticationBox.sharedCache` finally, as the reason for implementing this has been the incorrect relationships we're keeping in the Database. But I've looked at the code and now I'm uncertain wether the better point in time wouldn't be once `MastodonUser` has been cleared from CoreData.

PR #1143, opened 2023-11-04
Update design for verified links in profile (IOS-130)

zeitschlag:
As part of the profile-redesign, we wanted to change the "Verified"-badge from a simple checkmark to a ✨ sealed checkmark ✨ . And this does this PR.

On the left screenshot, you see how it was (green background, simple checkmark), on the two right screenshots, you see the updated design:

IOS-130

Please note the different images for darkmode/lightmode (thus: two screenshots), colors and fonts got slightly changed as well.


kimar (approved):
LGTM 🚀

PR #1149, opened 2023-11-09
Make "Followed Hashtags"-screen work with entities (IOS-186)

zeitschlag:
- Refactor "Followed Hashtags"-screen, it works with `Mastodon.Entity.Tag`
- Add "Pull to refresh"

------

Note: In case someone follows more than 100 hashtags, we should add pagination (it's been there since Mastodon `4.1`, see #1152)


kimar (approved):
LGTM 🚀

PR #1155, opened 2023-11-13
Make "Suggestions" use Entities (IOS-190)

zeitschlag:
- Replace Core Data on "Suggested Accounts"-screen which appears when you don't follow any people (For testing, you might need to create a new account)
- I also added some quality of life-improvements (like showing the spinner on every `UIViewController` and not just `rootViewController`)


MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
```diff
@@ -84,6 +84,64 @@ extension Mastodon.Entity {
...
+        hasher.combine(bot)
+        hasher.combine(source)
+        hasher.combine(suspended)
+        hasher.combine(muteExpiresAt)
+    }
+}
+
+extension Mastodon.Entity.Account: Equatable {
+    public static func == (lhs: Mastodon.Entity.Account, rhs: Mastodon.Entity.Account) -> Bool {
+        return lhs.id == rhs.id &&
```
zeitschlag: @kimar Stupid question: Can we compare `hashValue` here instead of comparing every property?

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
```diff
@@ -84,6 +84,64 @@ extension Mastodon.Entity {
     }
 }
 
+extension Mastodon.Entity.Account: Hashable {
```
DK1DA, line 89: Apparently there is already a conformance of this typ to `Hashable` and it's only taking into account the `id` ...
 Bildschirmfoto 2023-11-14 um 11 53 07

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
```diff
@@ -84,6 +84,64 @@ extension Mastodon.Entity {
...
+    public func hash(into hasher: inout Hasher) {
```
DK1DA, line 90: I'm wondering if taking properties such as `followersCount` et. al. might lead to differences on the same account because of stale local data. Maybe `id`, `username`, `acct`, `url` are enough?

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
DK1DA: Hm, I would have guessed so, but after reading https://developer.apple.com/documentation/swift/hashable/hashvalue, specifically `Hash values are not guaranteed to be equal across different executions of your program.`. I'm not so sure anymore wether we should use it...

kimar (approved):
LGTM in general, just a few remarks and ❓

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
zeitschlag: Oh, I didn't see that 🙈

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
zeitschlag: This is a good, valid point. What makes things even worse is that depending on the server you ask, the same user has a different ID. Take [me on mastodon.social](https://mastodon.social/api/v1/accounts/lookup?acct=zeitschlag@chaos.social) and [me on chaos.social](https://chaos.social/api/v1/accounts/lookup?acct=zeitschlag@chaos.social):

Screenshot 2023-11-14 at 12 06 05

So maybe we'd have to rely on `url` only?

DK1DA: Yes, I think we had a method which always gave us the full account handle (incl. server) I think we should compare those and I wouldn't take the other attributes into consideration.

zeitschlag: You mean `acctWithDomainIfMissing(_:)`, right? That still uses the local domain (and you'd need to bring the local domain in `==` or `hash_into`. Luckily, there's a computed property `domain` which uses `url`, so I decided to compare `url` directly in both `Hashable` and `Equatable`.

PR #1167, opened 2023-11-16
Remove Status from CoreData

kimar:
# Rationale

We've made the decision to radically decrease usage of Core Data as with the current implementation we're storing an excessive amount of data which leads to degraded performance and also invalid data over time, e.g. when account relations are changing but are not reflected in the App.

This PR removes usage of the `CoreDataStack.Status` entity and replaces it by `MastodonStatus`.


Mastodon/Persistence/FileManager+SearchHistory.swift
```diff
@@ -68,8 +68,12 @@ extension FileManager {
     }
 }
 
-extension FileManager {
-    public var documentsDirectory: URL? {
-        return self.urls(for: .documentDirectory, in: .userDomainMask).first
+public extension FileManager {
```
zeitschlag, line 71: What do you think about moving this into its own file as we will use these `documentsDirectory` in several files?

(There's no need to do anything from your side, I'd take care of it in #1192, I just want to hear your thoughts :))

Mastodon/Protocol/Provider/DataSourceFacade+Bookmark.swift
```diff
@@ -9,18 +9,24 @@ import UIKit
 import CoreData
 import CoreDataStack
 import MastodonCore
+import MastodonSDK
 
 extension DataSourceFacade {
     public static func responseToStatusBookmarkAction(
-        provider: UIViewController & NeedsDependency & AuthContextProvider,
-        status: ManagedObjectRecord<Status>
+        provider: UIViewController & NeedsDependency & AuthContextProvider & DataSourceProvider,
```
zeitschlag: You could use `ViewControllerWithDependencies` here for the first two protocols in case you need a `UIViewController` here which doesn't look like to me? oO

Mastodon/Protocol/Provider/DataSourceFacade+Follow.swift
```diff
@@ -93,22 +86,12 @@ extension DataSourceFacade {
             )
         } catch {
             // reset state when failure
-            try? await managedObjectContext.performChanges {
-                guard let notification = notification.object(in: managedObjectContext) else { return }
-                notification.transientFollowRequestState = .init(state: .none)
-            }
-
+            notification.transientFollowRequestState = .init(state: .none)
```
zeitschlag, line 89: This looks so much better without `try? await managedObjectContext.performChanges {...}`

Mastodon/Protocol/Provider/DataSourceFacade+Follow.swift
```diff
@@ -124,32 +107,14 @@ extension DataSourceFacade {
...
-            }
+        switch query {
+        case .accept:
+            notification.transientFollowRequestState = .init(state: .isAccept)
+            notification.followRequestState = .init(state: .isAccept)
+        case .reject:
+            break
         }
-    }   // end func
+    }
```
zeitschlag, line 117: 👍

MastodonSDK/Sources/MastodonCore/Persistence/Persistence.swift
```diff
@@ -10,11 +10,20 @@ import Foundation
...
     case searchHistory
+    case homeTimeline(String)
+    case notificationsMentions(String)
+    case notificationsAll(String)
 
     private var filename: String {
         switch self {
             case .searchHistory:
                 return "search_history"
+            case let .homeTimeline(userId):
```
zeitschlag: I know, what are the odds, but if someone has two accounts on two different servers with the same userID (it's not global, right?), there would be a collision and the two different timelines would share one JSON-file. Is this an issue?

(This is mostly a theoretic question, I'll have the same problem with `Mastodon.Entity.Account`.)

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Bookmark.swift
```diff
@@ -111,35 +72,7 @@ extension APIService {
...
-                        networkDate: response.networkDate
-                    )
-                )
-                
-                result.status.update(bookmarked: true, by: me)
-                result.status.reblog?.update(bookmarked: true, by: me)
-            }   // end for … in
-        }
-        
+
```
zeitschlag, line 75: 😍

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Favorite.swift
```diff
@@ -14,42 +14,26 @@ import CoreDataStack
 extension APIService {
     
     private struct MastodonFavoriteContext {
-        let statusID: Status.ID
+        let statusID: MastodonStatus.ID
```
zeitschlag, line 17: I replaced all those `MastodonUser.ID` with `String`. One less dependency and I don't think that they will change in the future. What do you think about this?

MastodonSDK/Sources/MastodonCore/Service/API/APIService+HashtagTimeline.swift
```diff
@@ -42,23 +42,18 @@ extension APIService {
...
-                        networkDate: response.networkDate
-                    )
-                )
-            }
+           for entity in response.value {
+               guard let poll = entity.poll else { continue }
+               _ = Persistence.Poll.createOrMerge(
+                in: managedObjectContext,
+                context: .init(domain: domain, entity: poll, me: me, networkDate: response.networkDate)
+               )
```
zeitschlag:
```suggestion
               _ = Persistence.Poll.createOrMerge(
                   in: managedObjectContext,
                   context: .init(domain: domain, entity: poll, me: me, networkDate: response.networkDate)
               )
```

MastodonSDK/Sources/MastodonCore/Service/API/APIService+HomeTimeline.swift
```diff
@@ -41,8 +41,20 @@ extension APIService {
...
         let managedObjectContext = self.backgroundManagedObjectContext
+        try await managedObjectContext.performChanges {
+           let me = authenticationBox.authentication.user(in: managedObjectContext)
 
+           for entity in response.value {
+               guard let poll = entity.poll else { continue }
+               _ = Persistence.Poll.createOrMerge(
+                in: managedObjectContext,
+                context: .init(domain: domain, entity: poll, me: me, networkDate: response.networkDate)
+               )
```
zeitschlag:
```suggestion
               _ = Persistence.Poll.createOrMerge(
                   in: managedObjectContext,
                   context: .init(domain: domain, entity: poll, me: me, networkDate: response.networkDate)
               )
```

MastodonSDK/Sources/MastodonCore/Service/API/APIService+PublicTimeline.swift
```diff
@@ -27,24 +27,20 @@ extension APIService {
...
-                )
-            }
+           let me = authenticationBox.authentication.user(in: managedObjectContext)
+
+           for entity in response.value {
+               guard let poll = entity.poll else { continue }
+               _ = Persistence.Poll.createOrMerge(
+                in: managedObjectContext,
+                context: .init(domain: domain, entity: poll, me: me, networkDate: response.networkDate)
+               )
```
zeitschlag:
```suggestion
               guard let poll = entity.poll else { continue }
               _ = Persistence.Poll.createOrMerge(
                   in: managedObjectContext,
                   context: .init(domain: domain, entity: poll, me: me, networkDate: response.networkDate)
               )
```

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Status.swift
```diff
@@ -133,3 +133,25 @@ extension Mastodon.Entity.Status {
...
+
+extension Mastodon.Entity.Status: Hashable {
+    public static func == (lhs: Mastodon.Entity.Status, rhs: Mastodon.Entity.Status) -> Bool {
+        lhs.uri == rhs.uri &&
+        lhs.id == rhs.id &&
+        lhs.reblog == rhs.reblog &&
+        lhs.favourited == rhs.favourited &&
+        lhs.reblogged == rhs.reblogged &&
+        lhs.bookmarked == rhs.bookmarked &&
+        lhs.pinned == rhs.pinned
```
zeitschlag, line 145: Just a question:

Why are those properties relevant for `Hashable`/`Equatable`? In case I favorited/pinned/reblogged/bookmarked a Status it count's as a new one?

MastodonSDK/Sources/MastodonSDK/MastodonFeed.swift
```diff
@@ -0,0 +1,75 @@
+// Copyright © 2023 Mastodon gGmbH. All rights reserved.
+
+import Foundation
+import CoreDataStack
```
zeitschlag, line 4: The `import`-statement is here for `Feed.Kind`, right?

MastodonSDK/Sources/MastodonSDK/MastodonFeed.swift
```diff
@@ -0,0 +1,75 @@
...
+
+public final class MastodonFeed {
```
zeitschlag, line 6: I haven't fully understood what we need `MastodonFeed` for. What is it good for?

MastodonSDK/Sources/MastodonSDK/MastodonNotification.swift
```diff
@@ -0,0 +1,48 @@
...
+    public init(entity: Mastodon.Entity.Notification, account: MastodonUser, status: MastodonStatus?, feeds: [MastodonFeed]) {
+        self.entity = entity
+        self.account = account
+        self.status = status
+        self.feeds = feeds
+    }
+}
+
+public extension MastodonNotification {
+    static func fromEntity(_ entity: Mastodon.Entity.Notification, using managedObjectContext: NSManagedObjectContext, domain: String) -> MastodonNotification? {
```
zeitschlag, line 29: I'm so looking forward to removing `managedObjectContext` and getting an `Mastodon.Entity.Account` here.

MastodonSDK/Sources/MastodonSDK/MastodonStatus.swift
```diff
@@ -0,0 +1,60 @@
...
+        hasher.combine(isSensitiveToggled)
+    }
+}
+
+public extension Mastodon.Entity.Status {
+    var asMastodonStatus: MastodonStatus {
+        .fromEntity(self)
+    }
+    
+    var mastodonVisibility: MastodonVisibility? {
```
zeitschlag, line 56: I have no idea, why this `MastodonVisibility` is still part of `CoreDataStack`?

MastodonSDK/Sources/MastodonUI/View/Content/MediaView+Configuration.swift
```diff
@@ -191,8 +191,8 @@ extension MediaView {
             )
         }
         
-        let status: StatusCompatible = status.reblog ?? status
-        let attachments = status.attachments
+//        let status: StatusCompatible = status.reblog ?? status
```
zeitschlag: What's the reason this is commented out?

MastodonSDK/Sources/MastodonUI/View/Content/StatusCardControl.swift
```diff
@@ -333,6 +338,18 @@ private extension Card {
...
 
+private extension Mastodon.Entity.Card {
+    var layout: StatusCardControl.Layout {
+        var aspectRatio = CGFloat(width ?? 1) / CGFloat(height ?? 1)
+        if !aspectRatio.isFinite {
+            aspectRatio = 1
+        }
+        return (abs(aspectRatio - 1) < 0.05 || image == nil) && html == nil
+        ? .compact
+        : .large(aspectRatio: aspectRatio)
```
zeitschlag: What about a proper `if else` here?

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -595,3 +533,15 @@ extension StatusView {
     }
 
 }
+
+extension MastodonStatus {
```
zeitschlag: Why is this extension in `StatusView+Configuration.swift`?

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -405,146 +387,102 @@ extension StatusView {
...
-                return bookmarkedBy.contains(where: {
-                    $0.id == authContext.mastodonAuthenticationBox.userID && $0.domain == authContext.mastodonAuthenticationBox.domain
-                })
-            }
-            .assign(to: \.isBookmark, on: viewModel)
-            .store(in: &disposeBag)
+        // relationship
+        viewModel.isReblog = status.entity.reblogged == true
+        viewModel.isFavorite = status.entity.favourited == true
+        viewModel.isBookmark = status.entity.bookmarked == true
```
zeitschlag, line 481: Why are these ` == true` necessary?

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -405,146 +387,102 @@ extension StatusView {
         pollTableViewDiffableDataSource?.applySnapshotUsingReloadData(_snapshot)
     }
 
-    private func configurePoll(status: Status) {
+    private func configurePoll(status: MastodonStatus) {
         let status = status.reblog ?? status
+
+        let predicate = Poll.predicate(domain: viewModel.authContext?.mastodonAuthenticationBox.domain ?? "", id: status.entity.poll?.id ?? "")
```
zeitschlag: Can you move those optionals into a `guard`? Does it make sense here?

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -178,109 +193,79 @@ extension StatusView {
...
+                    let metaContent = try MastodonMetaContent.convert(document: content)
+                    return metaContent
+                } catch {
+                    assertionFailure(error.localizedDescription)
+                    return PlaintextMetaContent(string: author.displayNameWithFallback)
+                }
+            }()
+            
+            // author username
+            viewModel.authorUsername = author.acct
```
zeitschlag, line 223: This looks so much better than the `.assign(to:)`-dance :)

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -178,109 +193,79 @@ extension StatusView {
                             }
                             .store(in: &disposeBag)
                     }   // end if let
-                }   // end else B2.
+//                }   // end else B2.
```
zeitschlag: ??

MastodonSDK/Sources/MastodonUI/Protocol/StatusCompatible.swift
```diff
@@ -2,6 +2,7 @@
 
 import Foundation
 import CoreDataStack
+import MastodonSDK
 
 public protocol StatusCompatible {
     var reblog: Status? { get }
```
zeitschlag, line 8: this `Status` comes from Core Data, right?

And `MastodonAttachment` are for another day, right?

Mastodon/Diffable/Report/ReportSection.swift
```diff
@@ -45,18 +45,15 @@ extension ReportSection {
...
-                        cell: cell,
-                        viewModel: .init(value: status),
-                        configuration: configuration
-                    )
-                }
+                configure(
+                    context: context,
+                    tableView: tableView,
+                    cell: cell,
+                    viewModel: .init(value: status),
```
zeitschlag, line 54: I just don't get why we have a `ViewModel` with exactly one property here. This overly complex layers mess with my head quite often. Also applies to `Configuration`.

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -100,11 +112,11 @@ extension DataSourceFacade {
     
         let managedObjectContext = provider.context.managedObjectContext
         let mentions = try? await managedObjectContext.perform {
```
zeitschlag: Why do we need this `managedObjectContext.perform`?

Mastodon/Protocol/Provider/DataSourceProvider.swift
```diff
@@ -39,4 +39,6 @@ extension DataSourceItem {
 
 protocol DataSourceProvider: ViewControllerWithDependencies {
     func item(from source: DataSourceItem.Source) async -> DataSourceItem?
+    func update(status: MastodonStatus)
+    func delete(status: MastodonStatus)
```
zeitschlag, line 43: What do you think about empty Default-implementations? I'm asking for

```swift
func update(status: MastodonStatus) {
    assertionFailure("Not required")
}
    
func delete(status: MastodonStatus) {
    assertionFailure("Not required")
}
```

Mastodon/Scene/Discovery/Posts/DiscoveryPostsViewModel.swift
```diff
@@ -41,14 +41,11 @@ final class DiscoveryPostsViewModel {
...
+    @MainActor
     init(context: AppContext, authContext: AuthContext) {
         self.context = context
         self.authContext = authContext
-        self.statusFetchedResultsController = StatusFetchedResultsController(
-            managedObjectContext: context.managedObjectContext,
-            domain: authContext.mastodonAuthenticationBox.domain,
-            additionalTweetPredicate: nil
-        )
+        self.statusFetchedResultsController = StatusFetchedResultsController()
```
zeitschlag: As someone who just removed `UserFetchedResultsController` and replaced it with `[Mastodon.Entity.Account]` and `[Mastodon.Entity.Relationship]`: Is there a reason to keep those `*FetchedResultControllers` and not replace it with an array of entities?

Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DataSourceProvider.swift
```diff
@@ -20,22 +21,28 @@ extension HomeTimelineViewController: DataSourceProvider {
         }
         
         switch item {
-        case .feed(let record):
-            let managedObjectContext = context.managedObjectContext
-            let item: DataSourceItem? = try? await managedObjectContext.perform {
-                guard let feed = record.object(in: managedObjectContext) else { return nil }
+        case .feed(let feed):
+            let item: DataSourceItem? = {
```
zeitschlag: What's the reason for this computed property?

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -15,20 +15,31 @@ extension DataSourceFacade {
...
-            target: target
-        )
+        let acct: String = {
+            switch target {
+            case .status:
+                return status.reblog?.entity.account.acct ?? status.entity.account.acct
+            case .reblog:
+                return status.entity.account.acct
+            }
+        }()
```
zeitschlag: What's the reason to use a function here? Why not something like this:

```swift
let acct: String
switch target {
  case .status:
    acct = status.reblog....
  case .reblog:
    acct = status.entity....
```

?

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -15,20 +15,31 @@ extension DataSourceFacade {
...
+        
+        let _redirectRecord = try? await Mastodon.API.Account.lookupAccount(
```
zeitschlag, line 31: IIRC we do use the spinner in `coordinateToScene` when fetching the user. Maybe it's an idea to use that here as well as `lookup` might take some time?

Mastodon/Protocol/Provider/DataSourceFacade+URL.swift
```diff
@@ -9,11 +9,12 @@ import Foundation
 import CoreDataStack
 import MetaTextKit
 import MastodonCore
+import MastodonSDK
 
 extension DataSourceFacade {
     static func responseToURLAction(
         provider: DataSourceProvider & AuthContextProvider,
-        status: ManagedObjectRecord<Status>,
+        status: MastodonStatus,
```
zeitschlag: Why do we need this `status` here, what do we use it for? oO

Mastodon/Protocol/Provider/DataSourceProvider+NotificationTableViewCellDelegate.swift
```diff
@@ -180,16 +179,13 @@ extension NotificationTableViewCellDelegate where Self: DataSourceProvider & Med
...
-                guard let notification = record.object(in: managedObjectContext) else { return nil }
-                guard let _status = notification.status else { return nil }
-                let status = _status.reblog ?? _status
+            let _mediaTransitionContext: NotificationMediaTransitionContext? = {
+                guard let status = record.status?.reblog ?? record.status else { return nil }
                 return NotificationMediaTransitionContext(
-                    status: .init(objectID: status.objectID),
-                    needsToggleMediaSensitive: status.isSensitiveToggled ? !status.sensitive : status.sensitive
+                    status: status,
+                    needsToggleMediaSensitive: status.isSensitiveToggled ? !(status.entity.sensitive == true) : (status.entity.sensitive == true)
```
zeitschlag: Maybe it's just me, but `?:` just got more complicated here :D is `== true` necessary? Would it be an idea to extract this to its own line?

Mastodon/Protocol/Provider/DataSourceProvider+StatusTableViewControllerNavigateable.swift
```diff
@@ -64,12 +65,7 @@ extension StatusTableViewControllerNavigateableCore where Self: DataSourceProvid
...
-            let _statusRecord: ManagedObjectRecord<Status>? = try? await context.managedObjectContext.perform {
-                guard let notification = record.object(in: self.context.managedObjectContext) else { return nil }
-                guard let status = notification.status else { return nil }
-                return .init(objectID: status.objectID)
-            }
-            guard let statusRecord = _statusRecord else {
+            guard let statusRecord = record.status else {
                 return nil
             }
             return statusRecord
```
zeitschlag: What about `return record.status` here?

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadOldestState.swift
```diff
@@ -52,13 +52,10 @@ extension HomeTimelineViewModel.LoadOldestState {
             }
             
             Task {
-                let managedObjectContext = viewModel.fetchedResultsController.managedObjectContext
-                let _maxID: Mastodon.Entity.Status.ID? = try await managedObjectContext.perform {
-                    guard let feed = lastFeedRecord.object(in: managedObjectContext),
-                          let status = feed.status
-                    else { return nil }
+                let _maxID: Mastodon.Entity.Status.ID? = {
```
zeitschlag: What do you think about something like this:
```swift
let _maxID = lastFeedRecord.status?.id
```

Mastodon/Scene/Notification/NotificationTimeline/NotificationTimelineViewController+DataSourceProvider.swift
```diff
@@ -20,22 +21,29 @@ extension NotificationTimelineViewController: DataSourceProvider {
...
             return nil
         }
     }
+
+    func update(status: MastodonStatus) {
+        viewModel.feedFetchedResultsController.update(status: status)
+    }
+    
+    func delete(status: MastodonStatus) {
+        viewModel.feedFetchedResultsController
```
zeitschlag: To me this looks a bit ... incomplete, is this intended?

Mastodon/Scene/Notification/NotificationTimeline/NotificationTimelineViewModel.swift
```diff
@@ -111,44 +99,35 @@ extension NotificationTimelineViewModel {
...
+//        guard let maxID = record.notification?.id else { return }
+
+//        // fetch data
+//        if let notifications = try? await context.apiService.notifications(
+//            maxID: maxID,
+//            scope: scope,
+//            authenticationBox: authContext.mastodonAuthenticationBox
+//        ) {
+//            self.feedFetchedResultsController.records += notifications.value.map { MastodonFeed.fromNotification($0, kind: record.kind) }
+//        }
```
zeitschlag: Can this go?

Mastodon/Scene/Report/Share/Cell/ReportStatusTableViewCell+ViewModel.swift
```diff
@@ -6,13 +6,13 @@
 //
 
 import UIKit
-import CoreDataStack
+import MastodonSDK
 
 extension ReportStatusTableViewCell {
     final class ViewModel {
```
zeitschlag, line 13: These viewModels with just one property don't make any sense to me 🙈

Mastodon/Scene/Root/MainTab/MainTabBarController.swift
```diff
@@ -170,145 +171,152 @@ extension MainTabBarController {
...
+            $avatarURL
+                .receive(on: DispatchQueue.main)
+                .sink { [weak self] avatarURL in
+                    guard let self = self else { return }
+                    self.avatarButton.avatarImageView.setImage(
+                        url: avatarURL,
+                        placeholder: .placeholder(color: .systemFill),
+                        scaleToSize: MainTabBarController.avatarButtonSize
+                    )
+                    //                self.avatarButton.avatarImageView.image = UIImage(named: "AppIcon")
```
zeitschlag: What do we need this for?

Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
```diff
@@ -123,6 +123,8 @@ extension SearchResultViewModel.State {
                     // discard result when state is not Loading
                     guard stateMachine.currentState is Loading else { return }
 
+                    let statusIDs = searchResults.statuses.map { MastodonStatus.fromEntity($0) }
```
zeitschlag: These aren't IDs anymore, right? 🙈

Mastodon/Scene/Thread/ThreadViewController+DataSourceProvider.swift
```diff
@@ -28,8 +29,121 @@ extension ThreadViewController: DataSourceProvider {
...
+        viewModel.mastodonStatusThreadViewModel.descendants.handleDelete(status: status, for: viewModel)
+    }
+    
     @MainActor
     private func indexPath(for cell: UITableViewCell) async -> IndexPath? {
         return tableView.indexPath(for: cell)
     }
 }
+
+private extension [StatusItem] {
```
zeitschlag, line 77: I didn't expect this :D

Mastodon/Scene/Thread/ThreadViewModel+LoadThreadState.swift
```diff
@@ -115,8 +115,10 @@ extension ThreadViewModel.LoadThreadState {
     
     class NoMore: ThreadViewModel.LoadThreadState {
         override func isValidNextState(_ stateClass: AnyClass) -> Bool {
-            return false
+            switch stateClass {
+            case is Loading.Type: return true
```
zeitschlag: May I suggest an `if` here?

MastodonSDK/Sources/MastodonCore/FetchedResultsController/FeedFetchedResultsController.swift
```diff
@@ -9,87 +9,86 @@
...
 
-// MARK: - NSFetchedResultsControllerDelegate
-extension FeedFetchedResultsController: NSFetchedResultsControllerDelegate {
-    public func controller(
-        _ controller: NSFetchedResultsController<NSFetchRequestResult>,
-        didChangeContentWith snapshot: NSDiffableDataSourceSnapshotReference
-    ) {
-        let snapshot = snapshot as NSDiffableDataSourceSnapshot<String, NSManagedObjectID>
-        self._objectIDs.send(snapshot.itemIdentifiers)
+private extension FeedFetchedResultsController {
```
zeitschlag: I don't know yet if I'm convinced, but I kinda like this `private extension`-approach 👍

zeitschlag (commented):
Impressive PR, thank you. There are some interesting ideas and from what I saw, it already looks really solid. Great work! 🚀 

For now I just went through the code and added some questions here and there. What I'll do next is to test it thoroughly on device (that's why it's not an *Approve* yet) and see how it works.

After that, I'd like to do a second round of Code Review, mainly to see if I missed something (I'm pretty sure I missed something :D)

Is that fine for you?

zeitschlag:
# Testing results (WIP)

I just needed a place to write down findings so consider this a breathing list:

- [x] When working with bookmarks (super useful for development!), the app crashes in [`Mastodon/DataSourceProvider+StatusTableViewCellDelegate.swift` in line 483](https://github.com/mastodon/mastodon-ios/blob/remove_status/Mastodon/Protocol/Provider/DataSourceProvider%2BStatusTableViewCellDelegate.swift#L483) due to a missing author
- [x] When I favorite a retooted status, the "Retooted by"-message disappears

Mastodon/Persistence/FileManager+SearchHistory.swift
DK1DA: Yes definitely sounds like a good idea to me 👍

MastodonSDK/Sources/MastodonCore/Persistence/Persistence.swift
DK1DA: Yes this would indeed be an issue and needs fixing, I'll take care of this to also consider the domain.

Mastodon/Scene/Root/MainTab/MainTabBarController.swift
DK1DA: Woopsie, this was part of debugging when trying to figure out why the avatar won't load anymore. I'm removing this.

Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
DK1DA: Absolutely! Good find 😄

Mastodon/Scene/Thread/ThreadViewModel+LoadThreadState.swift
DK1DA: What if we do neither...

```
stateClass is Loading.Type
```

MastodonSDK/Sources/MastodonCore/FetchedResultsController/FeedFetchedResultsController.swift
DK1DA: We could also make it a private function inside the owning class, but this it's a bit more structure IMO.

kimar:
> # Testing results (WIP)
> I just needed a place to write down findings so consider this a breathing list:
> 
> * [ ]  When working with bookmarks (super useful for development!), the app crashes in [`Mastodon/DataSourceProvider+StatusTableViewCellDelegate.swift` in line 483](https://github.com/mastodon/mastodon-ios/blob/remove_status/Mastodon/Protocol/Provider/DataSourceProvider%2BStatusTableViewCellDelegate.swift#L483) due to a missing author

Good that you noticed this! 

The crash happens because no user entity exists for this posts user anymore. 

I just realized that there are still a couple of places like this where the MastodonUser needs to be replaced, however I'm not sure if this conflicts with your user refactoring.

zeitschlag:
> however I'm not sure if this conflicts with your user refactoring.

I'm pretty sure it will, but so be it then. Also my domain-block-PR will create conflicts, but that's okay. I think it doesn't make sense to wait with those fixes until I'm finished with the `MastodonUser`-stuff

Mastodon/Protocol/Provider/DataSourceFacade+Status.swift
```diff
@@ -144,7 +144,8 @@ extension DataSourceFacade {
 extension DataSourceFacade {
     
     struct MenuContext {
-        let author: ManagedObjectRecord<MastodonUser>?
+        let author: ManagedObjectRecord<MastodonUser>? // todo: Remove once IOS-192 is ready
```
zeitschlag, line 147: Thank you for the comment 🫶

MastodonSDK/Sources/MastodonCore/Persistence/Persistence.swift
zeitschlag: TIL: `UserIdentifier`! Thank you :)

Mastodon/Protocol/Provider/DataSourceProvider+UITableViewDelegate.swift
```diff
@@ -42,22 +43,15 @@ extension UITableViewDelegate where Self: DataSourceProvider & AuthContextProvid
             case .notification(let notification):
                 let managedObjectContext = context.managedObjectContext
```
zeitschlag: What do we need this for?

MastodonSDK/Sources/MastodonCore/Persistence/Persistence.swift
```diff
@@ -10,20 +10,24 @@ import Foundation
...
         switch self {
             case .searchHistory:
-                return "search_history"
-            case let .homeTimeline(userId):
-                return "home_timeline_\(userId)"
-            case let .notificationsMentions(userId):
-                return "notifications_mentions_\(userId)"
-            case let .notificationsAll(userId):
-                return "notifications_all_\(userId)"
+                return "search_history" // todo: @zeitschlag should this be user-scoped as well?
```
zeitschlag, line 20: Very good point, thank you! I'll copy your `userIdentifier`-approach :)

MastodonSDK/Sources/MastodonCore/Persistence/Persistence.swift
```diff
@@ -10,20 +10,24 @@ import Foundation
...
-    case homeTimeline(String)
-    case notificationsMentions(String)
-    case notificationsAll(String)
+    case homeTimeline(UserIdentifier)
+    case notificationsMentions(UserIdentifier)
+    case notificationsAll(UserIdentifier)
+    
+    private func uniqueUserDomainIdentifier(for userIdentifier: UserIdentifier) -> String {
+        "\(userIdentifier.userID)@\(userIdentifier.domain)"
+    }
```
zeitschlag: We could put this as a default-implementation into the `UserIdentifier`-protocol?

Something tells me that that might come in very handy!

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Follow.swift
```diff
@@ -195,4 +195,38 @@ extension APIService {
...
+                domain: authenticationBox.domain,
+                accountID: user.id,
+                followQueryType: .follow(query: .init(reblogs: newShowReblogs)),
+                authorization: authenticationBox.userAuthorization
+            ).singleOutput()
+
+            result = .success(response)
+        } catch {
+            result = .failure(error)
+        }
```
zeitschlag, line 228: Wait a second: I know this code! :P

zeitschlag (commented):
Your changes since the last review look good, I like the `UserIdentifier`-approach. Can you say something about the other comments? Feel free to quietly resolve them :)

Mastodon/Protocol/Provider/DataSourceProvider+UITableViewDelegate.swift
DK1DA: Looks like a left-over, I'll remove it, thx for the hint and sorry for forgetting about this. It's just such a huge changelog 😮‍💨

MastodonSDK/Sources/MastodonCore/Persistence/Persistence.swift
DK1DA: Yes, very good point!

Mastodon/Protocol/Provider/DataSourceProvider+StatusTableViewControllerNavigateable.swift
DK1DA: Yep makes sense 👍

Mastodon/Protocol/Provider/DataSourceProvider.swift
DK1DA: I deliberately chose to not do this to make sure the implementation isn't silently forgotten somewhere.

Mastodon/Scene/Discovery/Posts/DiscoveryPostsViewModel.swift
DK1DA: I think some of them are shared IIRC, but let me double check 👀

Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DataSourceProvider.swift
DK1DA: Yeah, doesn't make much sense right? Aaaand it's gone.

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadOldestState.swift
DK1DA: Yep 👍

Mastodon/Scene/Notification/NotificationTimeline/NotificationTimelineViewController+DataSourceProvider.swift
DK1DA: Woopsie, no it's not intended but I think this code also isn't reachable, I'll fix this.

Mastodon/Scene/Report/Share/Cell/ReportStatusTableViewCell+ViewModel.swift
DK1DA: Yep I feel you, would make sense if we'd try to test it but this way it really just adds code, I'm adding a todo comment here but don't want to further expand the scope of this PR.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Favorite.swift
DK1DA: Hm I actually like this type alias, even though it's just shadowing `String` I feel like it adds a bit of documentation to the codebase.

MastodonSDK/Sources/MastodonCore/Service/API/APIService+PublicTimeline.swift
DK1DA: Hm weird auto indent puts the cursor where it is, but I'm manually indenting now for the sake of readability.

Mastodon/Protocol/Provider/DataSourceFacade+URL.swift
DK1DA: Hm weird, looks like it's been never used.

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
DK1DA: Removed.

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
DK1DA: Because `favourited`, `bookmarked` and `reblogged`.are optionals.

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -12,23 +12,37 @@ import MastodonSDK
...
+            query: .init(acct: acct),
+            authorization: provider.authContext.mastodonAuthenticationBox.userAuthorization
+        ).singleOutput().value
+                
         guard let redirectRecord = _redirectRecord else {
             assertionFailure()
+            provider.coordinator.hideLoading()
             return
         }
         await coordinateToProfileScene(
```
zeitschlag, line 44: I noticed that the spinner isn't hidden again, so maybe add a `provider.coordinator.hideLoading()` before coordinating to the profile-scene?

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Status.swift
DK1DA: Exactly that's the case! Otherwise the diffable datasource won't notice changes on the entities.

MastodonSDK/Sources/MastodonSDK/MastodonFeed.swift
DK1DA: Exactly!

MastodonSDK/Sources/MastodonSDK/MastodonFeed.swift
DK1DA: `MastodonFeed` replaces the original `Feed` item which is an abstraction over features such as loading gaps in between statuses. I _think_ we should be able to get rid of it, but I didn't dare to do so yet because it would have resulted in an even bigger set of changes.

MastodonSDK/Sources/MastodonSDK/MastodonNotification.swift
DK1DA: Yes, I totally feel you!

MastodonSDK/Sources/MastodonSDK/MastodonStatus.swift
DK1DA: Because I didn't remove the Status entities (yet) and it's still used there, that said I'M wondering if I should actually remove the whole `CoreDataStack.Status` model classes...

MastodonSDK/Sources/MastodonUI/Protocol/StatusCompatible.swift
DK1DA: 1. Yes
2. I guess we'll be able to remove it once we remove the old `Status`

MastodonSDK/Sources/MastodonUI/View/Content/MediaView+Configuration.swift
DK1DA: Ah forgot to remove this line, the `MastodonStatus` to be used will be determined by the caller and not inside of each individual `configuration(status:)` function.

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
DK1DA: Good point, I've moved it to `MastodonStatus`.

Mastodon/Scene/HashtagTimeline/HashtagTimelineViewModel.swift
```diff
@@ -24,7 +24,7 @@ final class HashtagTimelineViewModel {
     // input
     let context: AppContext
     let authContext: AuthContext
-    let fetchedResultsController: StatusFetchedResultsController
+    let fetchedResultsController: StatusDataController
```
zeitschlag:
```suggestion
    let dataController: StatusDataController
```

Mastodon/Scene/HashtagTimeline/HashtagTimelineViewModel.swift
```diff
@@ -50,15 +50,12 @@ final class HashtagTimelineViewModel {
...
     init(context: AppContext, authContext: AuthContext, hashtag: String) {
         self.context  = context
         self.authContext = authContext
         self.hashtag = hashtag
-        self.fetchedResultsController = StatusFetchedResultsController(
-            managedObjectContext: context.managedObjectContext,
-            domain: authContext.mastodonAuthenticationBox.domain,
-            additionalTweetPredicate: nil
-        )
+        self.fetchedResultsController = StatusDataController()
```
zeitschlag:
```suggestion
        self.dataController = StatusDataController()
```

Mastodon/Scene/HashtagTimeline/HashtagTimelineViewModel+State.swift
```diff
@@ -158,7 +159,7 @@ extension HashtagTimelineViewModel.State {
                         await enter(state: NoMore.self)
                     }
                     
-                    viewModel.fetchedResultsController.append(statusIDs: statusIDs)
+                    await viewModel.fetchedResultsController.setRecords(statusIDs.map { MastodonStatus.fromEntity($0) })
```
zeitschlag:
```suggestion
                    await viewModel.dataController.setRecords(statusIDs.map { MastodonStatus.fromEntity($0) })
```

Mastodon/Scene/HashtagTimeline/HashtagTimelineViewModel+State.swift
```diff
@@ -145,10 +146,10 @@ extension HashtagTimelineViewModel.State {
                     self.maxID = newMaxID
                     
                     var hasNewStatusesAppend = false
-                    var statusIDs = isReloading ? [] : viewModel.fetchedResultsController.statusIDs
+                    var statusIDs = isReloading ? [] : await viewModel.fetchedResultsController.records.map { $0.entity }
```
zeitschlag:
```suggestion
                    var statusIDs = isReloading ? [] : await viewModel.dataController.records.map { $0.entity }
```

Mastodon/Scene/HashtagTimeline/HashtagTimelineViewModel.swift
zeitschlag: I'm pretty sure I missed some `fetchedResultsController`s, but as I saw it being renamed to `dataController` somewhere else, so I thought: Why not be consistent? Let's be nice and use this suggestion-thing (which doesn't work for unchanged code?)

Mastodon/Scene/HomeTimeline/HomeTimelineViewController+DataSourceProvider.swift
```diff
@@ -20,22 +21,25 @@ extension HomeTimelineViewController: DataSourceProvider {
...
+                return nil
             }
-            return item
         default:
             return nil
         }
     }
+
+    func update(status: MastodonStatus) {
+        viewModel.fetchedResultsController.update(status: status)
```
zeitschlag: `s/fetchedResultsController/dataController/`?

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -114,6 +111,22 @@ extension HomeTimelineViewModel.LoadLatestState {
...
+                        for (i, record) in newRecords.enumerated() {
+                            if let index = oldRecords.firstIndex(where: { $0.status?.reblog?.id == record.id || $0.status?.id == record.id }) {
+                                oldRecords[index] = record
+                                if newRecords.count > index {
+                                    newRecords.remove(at: i)
+                                }
+                            }
+                        }
+                        return (newRecords + oldRecords).removingDuplicates()
+                    }()
```
zeitschlag, line 129: Just a question, if I understood this correctly:

This is bascially to replace updated statuses and append new ones, right?

Mastodon/Scene/Report/Share/Cell/ReportStatusTableViewCell+ViewModel.swift
```diff
@@ -6,13 +6,14 @@
 //
 
 import UIKit
-import CoreDataStack
+import MastodonSDK
 
 extension ReportStatusTableViewCell {
+    // todo: refactor / remove this
```
zeitschlag, line 12: What do you think about creating an issue for this? Could be a good first issue for someone :)

Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
```diff
@@ -147,21 +147,20 @@ extension SearchResultViewModel.State {
...
                     if offset == nil {
+                        await viewModel.dataController.reset()
                         viewModel.relationships = []
                         viewModel.accounts = []
-                        viewModel.statusFetchedResultsController.statusIDs = []
                         viewModel.hashtags = []
                     }
 
-                    // due to combine relationships must be set first
+                    await viewModel.dataController.appendRecords(statuses)
```
zeitschlag: Is there a reason you removed the comment? Did I understand something wrong and it wasn't even necessary in the first place?

Mastodon/Scene/Share/View/Content/NotificationView+Configuration.swift
```diff
@@ -16,30 +16,33 @@ import MastodonAsset
...
             return
         }
-        
-        configure(notification: notification)
+
+        MastodonNotification.fromEntity(
+            notification,
+            using: managedObjectContext,
+            domain: viewModel.authContext?.mastodonAuthenticationBox.domain ?? ""
+        ).map(configure(notification:))
```
zeitschlag, line 35: I never thought about using `map` for this

Mastodon/Scene/Thread/ThreadViewController.swift
```diff
@@ -182,3 +198,17 @@ extension ThreadViewController: StatusTableViewControllerNavigateable {
...
+            provider?.delete(status: status )
+        }
+    }
+    
+    func notifyChildrenAboutStatusUpdate(_ status: MastodonStatus) {
+        viewControllers.compactMap { $0 as? DataSourceProvider }.forEach { provider in
+            provider?.update(status: status )
+        }
+    }
+}
```
zeitschlag, line 214: Neat!

Mastodon/Scene/Thread/ThreadViewModel.swift
```diff
@@ -55,40 +58,25 @@ class ThreadViewModel {
...
             .sink { [weak self] root in
                 guard let self = self else { return }
                 guard case let .root(threadContext) = root else { return }
-                guard let status = threadContext.status.object(in: self.context.managedObjectContext) else { return }
+                let status = threadContext.status
                 
                 // bind threadContext
                 self.threadContext = .init(
-                    domain: status.domain,
+                    domain: authContext.mastodonAuthenticationBox.domain, //status.domain,
```
zeitschlag: I was curious why we need the domain here and it seems like we don't? What do you think?

MastodonSDK/Sources/MastodonCore/Service/API/APIService+Favorite.swift
zeitschlag: Good point

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
```diff
@@ -139,21 +148,27 @@ extension StatusView {
...
+                if let authenticationBox = viewModel.authContext?.mastodonAuthenticationBox {
+                    Task { @MainActor in
+                        if let replyTo = try? await Mastodon.API.Statuses.status(
+                            session: .shared,
+                            domain: authenticationBox.domain,
+                            statusID: inReplyToID,
+                            authorization: authenticationBox.userAuthorization
+                        ).singleOutput().value {
+                            let header = createHeader(name: replyTo.account.displayNameWithFallback, emojis: replyTo.account.emojis?.asDictionary ?? [:])
+                            viewModel.header = header
```
zeitschlag, line 166: Could it be that the header-reload-issue I sent you the video from gets triggered by this?

Mastodon/Scene/Root/MainTab/MainTabBarController.swift
```diff
@@ -170,145 +171,151 @@ extension MainTabBarController {
...
-        }
-        _viewControllers = viewControllers
-        setViewControllers(viewControllers, animated: false)
-        selectedIndex = 0
-
-        // hacky workaround for FB11986255 (Setting accessibilityUserInputLabels on a UITabBarItem has no effect)
-        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(50)) {
-            if let searchItem = self.tabBar.subviews.first(where: { $0.accessibilityLabel == Tab.search.title }) {
-                searchItem.accessibilityUserInputLabels = Tab.search.inputLabels
+        Task { @MainActor in
```
zeitschlag: Why is this (and the await in l. 179) needed?

`viewDidLoad` should run on `Main` nevertheless, why start a Task on `Main` to do `Main`-work? I'm sure: I'm missing something :)

zeitschlag (commented):
It gets better and better :) Great work!

I added some remarks and will do a test on device in the next few hours.

Mastodon/Scene/Thread/ThreadViewModel.swift
DK1DA: You're right, this was previously used when fetching the cached (domain scoped) statuses from Core Data. I'm removing this.

Mastodon/Scene/Root/MainTab/MainTabBarController.swift
DK1DA: That's because the method in [L:179](https://github.com/mastodon/mastodon-ios/pull/1167/files/c1ad3e2a9c7cabe81f99920b1bfda396d45a2a82#diff-4f9c4dc8589cb44d4bbf40a2b9ed397b53261d8995625b1b8e760d35341650d8R179) is an `async` method which as there _was_ a time (briefly while refactoring) when the `ProfileViewModel` did some async work in it's contractor. That being said, this wasn't necessary anymore and I've removed it now.

Very very good catch 🦅 👁️ !

Mastodon/Scene/Thread/ThreadViewController.swift
DK1DA: Thx!

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+Configuration.swift
zeitschlag: When I want to see sensitive content of a reblogged status, I also get this behaviour that the "Reblogged by" disappears

zeitschlag: We'll open an issue for this weird header-reload.

zeitschlag (approved):
LGTM, finally! 🚀 👍 🎊 🥳 

Thank you for that impressive amount of work.

Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
DK1DA: Hm good point actually, I'll need to check this 👀

Mastodon/Scene/Report/Share/Cell/ReportStatusTableViewCell+ViewModel.swift
DK1DA: Done: https://github.com/mastodon/mastodon-ios/issues/1206

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
DK1DA: Yes, it's supposed to replace existing status which are returned from the API because they might have changed and additionally to prepend new ones on the timeline.

Mastodon/Scene/Discovery/Posts/DiscoveryPostsViewModel.swift
DK1DA: As mentioned, the DataControllers are shared among other classes thus I'd like to keep them, but as you noticed I've renamed them properly.

Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
DK1DA: While I haven't spotted any issue with the changed order I'm still reverting and re-adding the comment here, to me it actually makes sense to have the relationships set upfront as setting the entities will lead to a refresh of the list.

zeitschlag: I think that was exactly the issue, IIRC. Thank you!

PR #1170, opened 2023-11-17
Replace Core Data on Suggested Accounts in Discovery-Section (IOS-194)

zeitschlag:
- Use `Mastodon.Entity.Account` instead of `MastodonUser` for suggested accounts incl. `Mastodon.Entity.Relationship` for those accounts
- Update Follow-button (it's blurple now instead of blacknwhite) including the Button-state
- Simplify things here and there

This is how it looks like now (for me):


Mastodon/Scene/Discovery/ForYou/DiscoveryForYouViewController.swift
```diff
@@ -117,17 +110,20 @@ extension DiscoveryForYouViewController: ProfileCardTableViewCellDelegate {
...
         Task {
-            try await DataSourceFacade.responseToUserFollowAction(
-                dependency: self,
-                user: record
-            )
-        }   // end Task
+            let newRelationship = try await DataSourceFacade.responseToUserFollowAction(dependency: self, user: account)
+
+            await MainActor.run {
+                cell.profileCardView.updateButtonState(with: newRelationship, isMe: false)
```
DK1DA: ❓  Are you sure there is no need to check for `account.id == authContext...user.id` for `isMe`?

kimar (approved):
Looks good to me 👍

Mastodon/Scene/Discovery/ForYou/DiscoveryForYouViewController.swift
zeitschlag: Honestly: No idea 😁 

My thoughts were: Why should my instance recommend me to myself? This would make any sense! But now I should think about it. Thank you for pointing this out.

PR #1171, opened 2023-11-18
Restore Credentials from Keychain

### LINKED ISSUE #1168: [BUG] When sharing article through share button in iOS, message “No available account”

### Current Behavior

When I try to share an article from a news app and select Mastodon, I see the message No Available Account. This happens both in iOS and iPadOS. I tried reinstalling the app and logging out and in again.

image

### Expected Behavior

When I share a article via a news app to the Mastodon app, I expect to see a title and a hyperlink, after which I can press Toot. The shared article should appear in my timeline.

### Steps To Reproduce

1. In a news app
2. Click the share icon
3. Choose Mastodon
4. A dialog appears with the text No Available Accounts 

### Environment

```markdown
- Device: iPhone 15/iPad 10
- OS: iOS/iPadOS
- Version: 17.0.3 
- Build: iOS/iPadOS 17.0.3 (21A360)
```

### Anything else?

_No response_

zeitschlag:
Seems to me that we overlooked this when migrating the credentials to keychain. At least we didn't restore credentials when we need them. I'm not sure how we can prevent this from happening in the future other than being and testing more thorough?

Closes #1168.

### DIFF
--- a/ShareActionExtension/Scene/ShareViewController.swift
+++ b/ShareActionExtension/Scene/ShareViewController.swift
@@ -160,9 +160,11 @@ extension ShareViewController {
 
 extension ShareViewController {
     private func setupAuthContext() throws -> AuthContext? {
-        let _authentication = AuthenticationServiceProvider.shared.authenticationSortedByActivation().first
-        let _authContext = _authentication.flatMap { AuthContext(authentication: $0) }
-        return _authContext
+        AuthenticationServiceProvider.shared.restore()
+
+        let authentication = AuthenticationServiceProvider.shared.authenticationSortedByActivation().first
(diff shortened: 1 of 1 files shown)


kimar (approved):
👍 thank you for fixing this. In this particular case (when migrating the credentials) we should have tested this 😞

zeitschlag:
@kimar Yes, I agree. But hey, this made us aware that we should test extensions.

A few years ago I was employed at a startup that built a medical device and those medical device regulations required a testplan to be written and passed before every single release, maybe we can think/talk about something like that (though much simpler) and even if the result is a short checklist what needs to be passed? If the consequence of that is a UI-test or a manual test, I don't care^^ I'd love to hear your thoughts

PR #1177, opened 2023-11-23
Remove CoreData for Tags/Users on Search

zeitschlag:
- To state the obvious: Don't use Core Data for Tags and Users on Search-screen
- Add a basic, JSON-based persistence (see `FileManager+SearchHistory.swift`)
- We throw away the (existing) SearchHistory.


kimar (approved):
LGTM in general but there are two things which look a bit odd to me:

1. When searching the search item is only added if interected with one of the results, e.g. when searching for a user it will only be persisted when viewing the user's profile. But this has already been the case before so I guess that's a design decision.
2. The results are only displayed upon re-entering the search results controller, e.g. but not immediately when returning from the search, maybe you could reload the saved items after a successful search / upon re-appearance of the search results VC ?

zeitschlag:
> 1. When searching the search item is only added if interected with one of the results, e.g. when searching for a user it will only be persisted when viewing the user's profile. But this has already been the case before so I guess that's a design decision.

AFAIK this is a design decision. We just save hashtags and profiles.

> 2. The results are only displayed upon re-entering the search results controller, e.g. but not immediately when returning from the search, maybe you could reload the saved items after a successful search / upon re-appearance of the search results VC ?

Will look into it, thank you!

PR #1179, opened 2023-11-23
Remove Cocoapods

zeitschlag:
Closes #1027


ahmdyasser (approved):

.github/scripts/setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# workaround https://github.com/CocoaPods/CocoaPods/issues/11355
-# sed -i '' $'1s/^/source "https:\\/\\/github.com\\/CocoaPods\\/Specs.git"\\\n\\\n/' Podfile
+brew install swiftgen
+brew install sourcery
```
DK1DA: I'm a huge fan of getting rid of CocoaPods **but**...

Some thoughts from my side:

1. We should definitely somehow pin the versions (e.g. `brew install swiftgen@x.y.z`)
2. I'm really not a fan of messing around with the user's system / installing our project dependencies system-wide as this might break setups on other ends. Maybe we can install swiftgen and sourcery somehow else? 

How about we download the binary ZIPs as part of the setup / bootstrap process and put them somewhere `.gitignore`d in our project?

Documentation/Setup.md
```diff
@@ -6,13 +6,20 @@
...
-
 Install the latest version of Xcode from the App Store or Apple Developer Download website. Also, we assert you have the [Homebrew](https://brew.sh) package manager.  
 
 This guide may not suit your machine and actually setup procedure may change in the future. Please file the issue or Pull Request if there are any problems.
 
-## CocoaPods
-The app use [CocoaPods]() and [Arkana](https://github.com/rogerluan/arkana). Ruby Gems are managed through Bundler. Make sure you have [Rosetta](https://support.apple.com/en-us/HT211861) installed if you are using the M1 Mac.
+### Swiftgen and Sourcery
+
+This app uses [SwiftGen](https://github.com/SwiftGen/SwiftGen) and [Sourcery](https://github.com/krzysztofzablocki/Sourcery) for Code Generation. Please insteall them
```
DK1DA: It's *free* software 😄 

```suggestion
This app uses [SwiftGen](https://github.com/SwiftGen/SwiftGen) and [Sourcery](https://github.com/krzysztofzablocki/Sourcery) for Code Generation. Please install them
```

.github/scripts/setup.sh
zeitschlag: Pinning the version is a good idea, will do that, thank you! 👍 (Hopefully we won't forget to update them 🥲 )

As this is the script for Github actions, I don't really care about the user's system at this point :)

Good thing is that both dependencies `swiftgen` and `sourcery` are optional: There's just a warning if you don't have them installed and prefer to write your `strings` by hand. In the end compare it to `rbenv` or `arkana`: It's a (handy) requirement from our side. So it's kind of "take it or leave it" — or find your own way of dealing with things :)

> as this might break

Have you experienced that before or is it just a gut feeling?

Adding another manual step to the setup-process is error-prone (or I'm just an idiot, but most probably not the only one :D). An alternative would be to include both binaries in the repository but to be honest: That doesn't sound very appealing either.

DK1DA: > Good thing is that both dependencies `swiftgen` and `sourcery` are optional

Ah nice, then let's maybe add a note in the README so people are at least aware that they're installing something system-wide

> Have you experienced that before or is it just a gut feeling?

Yep a dozen times with all sorts of dependencies, in iOS development e.g. when people don't use bundler but randomly mess up their (or others) environment by globally installing their gems and (potentially) overriding others.

>  An alternative would be to include both binaries in the repository

Yep, I wouldn't pull in random binaries here

zeitschlag: I added a colorful hint (TIL: [Alert](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)):

Screenshot 2024-01-03 at 15 35 02

kimar (approved):
Lgtm 🎉

Issue #1187, opened 2023-11-30
[Feature]: Search my posts by keywords


testuser51:
### Description

Ability to search your posts by keywords

### Suggested Solution

1. Open your profile
2. Press 3 dots
3. "search by keywords"

### Alternatives

Add hints from the web version to the search field
![image](https://github.com/mastodon/mastodon-ios/assets/53906414/a1cedbf8-8b87-4cac-afb9-b78209891f1b)

But their problem is that it's unclear what the mastodon wants: a username, a username with a server, or a direct link to the profile (a bit confusing). Mastodon does not offer to automatically complete the field when you start typing

zeitschlag:
@testuser51 When #1188 is merged you should be able to use the search options in the next release. So if you entered `from:testuser51` in the search field and press the `search`-button on the keyboard, you'd see your toots.

Does this solve your issue?

testuser51:
It's not a perfect solution, but it's okay for now. 

But it would be nice if the search worked with alternative options:

- from:@testuser51
- from:@testuser51@server.name
- from:testuser51

zeitschlag:
@testuser51 The app uses the same Search-API as the Web-app. So that should work.

PR #1188, opened 2023-11-30
Fix endless fail-loop in search

zeitschlag:
Connected to #1187

If no accounts could be found for your search query, `ListBatchViewModel` and the search for relationships lead to an infinite loop always resulting in a `Fail`-state, but retrying every second resulting in a `Fail`-state, but retrying every second resulting in a `Fail`-state, but retrying every second resulting in a `Fail`-state, but retrying every second resulting in a `Fail`-state, but retrying every second (you get the picture, right?)

### DIFF
--- a/Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
+++ b/Mastodon/Scene/Search/SearchDetail/SearchResult/SearchResultViewModel+State.swift
@@ -125,10 +125,15 @@ extension SearchResultViewModel.State {
 
                     let accounts = searchResults.accounts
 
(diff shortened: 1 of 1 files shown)


kimar (approved):
LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, LGTM, …

Issue #1178, opened 2023-11-23
[BUG] An offline account blocks refresh for all other accounts


vszakats:
### Is there an existing issue for this?

- [X] I have searched the issues (both open and closed)

### Current Behavior

Normally after opening the app, the feed refreshes. Or if not,
it does after a manual pull.

After a recent update, instead of refreshing, it says "Offline"
at the top of the display and the feed remains unchanged.

The app has multiple accounts configured. I noticed that one
of them (an unselected/inactive one) is offline. After logging
out, refresh works again.

~~This only works until re-launching the Mastodon app, when
the logged-out account unexpectedly reappears as logged-in
and blocking refresh again. See #1172.~~ [FIXED]

IMO one offline server should not block access to the rest
of the feeds from other servers.

Possibly related past (open) issue: #898

### Expected Behavior

Feed refresh works when one of the (inactive) accounts is offline.

### Steps To Reproduce

1. Have multiple logged-in accounts.
2. Select one where the server works correctly.
3. Have one other account where the server is offline / inaccessible.
4. Open app.
5. Wait or pull to refresh.
6. "Offline" appears instead of refreshing.

### Environment

```markdown
- Device: iPhone
- OS: iOS 17
- Version: 2013.15
```

zeitschlag:
Notes to self:

- Check `networkErrorCount` (and its publisher) in `HomeTimelineNavigationBarTitleViewModel` 
    - Seems to work as intended
- Offline-badge appears when three things go wrong on Hometimeline. And a lot can go wrong there:
    - Decoding fails aka invalid status from API (gut feeling says: probably this)
    - Network-Errors
    - `accountInfo` for all logged in accounts (WHY? -> For account-list)
    - Core Data (fetch me leads to a `assertionFailure()`, should only crash in `DEBUG`)
    - ListBatchView triggers once per second

zeitschlag:
@vszakats Can you tell me which instances you're using?

vszakats:
@zeitschlag: Two popular fediverse ones and an experimental locally hosted one (this one was down). After deleting the latter, the 'Offline' issues resolved right away and working without issues since.

Edit: I remember seeing 'Offline' also on the account page of one clearly online public account, while this issue was in effect.

zeitschlag:
@vszakats Which fediverse ones? If you don't wanna mention them here, feel free to send me a DM :arrow_right: https://chaos.social/@zeitschlag/ (or don't, that's fine with me, too :))

[...]

PR #1189, opened 2023-11-30
Don't throw when one account can't be authenticated (#1178)

zeitschlag:


MastodonSDK/Sources/MastodonCore/Service/API/APIService+HomeTimeline.swift
```diff
@@ -46,8 +46,9 @@ extension APIService {
         // FIXME: This is a dirty hack to make the performance-stuff work.
         // Problem is, that we don't persist the user on disk anymore. So we have to fetch
         // it when we need it to display on the home timeline.
+        // We need this (also) for the Account-list, but it might be the wrong place. App Startup might be more appropriate
         for authentication in AuthenticationServiceProvider.shared.authentications {
-            _ = try await accountInfo(domain: authentication.domain,
+            _ = try await? accountInfo(domain: authentication.domain,
```
DK1DA: Well that's an interesting `?` position you got there, I thought this ought to be `try? await` but here I stand corrected 😄

kimar (approved):

MastodonSDK/Sources/MastodonCore/Service/API/APIService+HomeTimeline.swift
zeitschlag: 🫣

let's pretend this never happened!

DK1DA: I literally thought this would work as well there. Haven't tried to compile it (yet).

PR #1192, opened 2023-12-07
Make Profile-screen use Mastodon.Entity.Account instead of MastodonUser

zeitschlag:
- Make Profile-screen use Mastodon.Entity.Account instead of MastodonUser
- Fix things left and right
- Just like with the Status-refactoring I'm expecting more bugs to show up over time.

<details>
<summary>Checklist</summary>
- [x] Sharing/Blocking/Muting/Domain Block
- [x] Follow/Unfollow, Request
- [x] Being Followed/Unfollowed
- [x] Polls
- [x] Notification (Avatar-Buttons, auch bei Quoted Posts)
- [x] Show/Hide sensitive content
- [x] Profile: Check Relationship-Button
- [x] Reporting
- [x] Compose/Answer
- [x] Retoot, Favorite, Delete
- [ ] ~~Widgets!~~ See #1234
- [x] ShareExtension
- [x] StatusEdits
</details>


Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -46,54 +46,83 @@ extension DataSourceFacade {
...
-            return
+        provider.coordinator.showLoading()
+
+        Task {
+            do {
+                guard let account = try await provider.context.apiService.fetchUser(username: username,
+                                                                                    domain: domain,
+                                                                                    authenticationBox: provider.authContext.mastodonAuthenticationBox) else {
+                    return provider.coordinator.hideLoading()
+                }
```
DK1DA: I know this Is probably a controversial thing, but I really think the following is much more readable (less indentations )

```suggestion
                guard let account = try await provider.context.apiService.fetchUser(
                    username: username,
                    domain: domain,
                    authenticationBox: provider.authContext.mastodonAuthenticationBox
                ) else {
                    return provider.coordinator.hideLoading()
                }
```

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -46,54 +46,83 @@ extension DataSourceFacade {
...
+
+                provider.coordinator.hideLoading()
+
+                await coordinateToProfileScene(provider: provider, account: account)
```
DK1DA: Apparently `coordinateToProfileScene(provider:account:)` doesn't need to be awaited for, maybe related to my change on Status?

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -46,54 +46,83 @@ extension DataSourceFacade {
...
+                    authorization: provider.authContext.mastodonAuthenticationBox.userAuthorization
+                ).value
+
                 provider.coordinator.hideLoading()
-                
-                if let user {
-                    await coordinateToProfileScene(provider: provider, user: user.asRecord)
-                }
+
+                await coordinateToProfileScene(provider: provider, account: account)
```
DK1DA, line 93: Same here regarding `await`, looks like it's not required here

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -46,54 +46,83 @@ extension DataSourceFacade {
...
                 provider.coordinator.hideLoading()
             }
         }
     }
+
+    @MainActor
+    static func coordinateToProfileScene(
+        provider: ViewControllerWithDependencies & AuthContextProvider,
+        account: Mastodon.Entity.Account
+    ) {
```
DK1DA: Hoe about making this function `async` ? Then you could get rid of the `Task` below, also the `Task { @MainActor ... ` together with scoping this method `@MainActor` feels a bit weird.

However if you make this method `async` you'll need to keep the `await`s above of course 😄

Mastodon/Protocol/Provider/DataSourceFacade+Status.swift
```diff
@@ -144,18 +144,18 @@ extension DataSourceFacade {
...
     }
     
     @MainActor
-    static func responseToMenuAction(
+    static func responseToMenuAction<T>(
         dependency: UIViewController & NeedsDependency & AuthContextProvider & DataSourceProvider,
         action: MastodonMenu.Action,
-        menuContext: MenuContext
+        menuContext: MenuContext,
+        completion: ((T) -> Void)? = { (param: Void) in }
```
DK1DA, line 151: In other functions, e.g `responseToUserFollowAction` we're simply returning the `Mastodon.Entity.Relationship`, why can't this be done here as well? I feel like another closure makes things more complex than it needs to be.

Mastodon/Protocol/Provider/DataSourceProvider+NotificationTableViewCellDelegate.swift
```diff
@@ -320,9 +304,11 @@ extension NotificationTableViewCellDelegate where Self: DataSourceProvider & Aut
                 return
             }
 
+            guard let account = notification.status?.entity.account else { return }
```
DK1DA, line 334: ❓  No directly related to your changes, but I'm wondering if maybe in such cases we should raise an `assertionFailure()` so we're spotting this in Debug builds? (I have also not yet done this, but maybe we should in the future?) WDYT?

Mastodon/Protocol/Provider/DataSourceProvider+NotificationTableViewCellDelegate.swift
```diff
@@ -494,21 +480,18 @@ extension NotificationTableViewCellDelegate where Self: DataSourceProvider & Aut
...
-            case .user(let user):
-                await DataSourceFacade.coordinateToProfileScene(
-                    provider: self,
-                    user: user
-                )
-            case .notification:
-                assertionFailure("TODO")
-            default:
-                assertionFailure("TODO")
+                case .status(let status):
```
DK1DA: I'm just wondering about the indentations here because for me, and I think with the Xcode default settings, the above is the correctly indented version.

Mastodon/Protocol/Provider/DataSourceProvider+StatusTableViewCellDelegate.swift
```diff
@@ -709,21 +682,21 @@ extension StatusTableViewCellDelegate where Self: DataSourceProvider & AuthConte
...
-            case .user(let user):
-                await DataSourceFacade.coordinateToProfileScene(
-                    provider: self,
-                    user: user
-                )
-            case .notification:
-                assertionFailure("TODO")
-            default:
-                assertionFailure("TODO")
+                case .status(let status):
```
DK1DA: Same as my comment regarding indentations above, re-indenting in Xcode it will produce changes here 🤔

Mastodon/Scene/Discovery/ForYou/ProfileCardView+Configuration.swift
```diff
@@ -22,7 +22,7 @@ extension ProfileCardView {
         viewModel.followersCount = account.followersCount
         viewModel.authorAvatarImageURL = account.avatarImageURL()
 
-        let emojis = account.emojis?.asDictionary ?? [:]
+        let emojis = account.emojis.asDictionary
```
DK1DA, line 25: 😍

Mastodon/Scene/Discovery/ForYou/ProfileCardView.swift
```diff
@@ -316,9 +316,9 @@ extension ProfileCardView {
                 buttonState = .none
             } else if relationship.following {
                 buttonState = .unfollow
-            } else if relationship.blocking || (relationship.domainBlocking ?? false) {
+            } else if relationship.blocking || relationship.domainBlocking {
                 buttonState = .blocked
-            } else if relationship.requested ?? false {
+            } else if relationship.requested {
```
DK1DA, line 321: 😍 😍

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -92,6 +92,7 @@ extension HomeTimelineViewModel.LoadLatestState {
             }
 
             do {
+                await AuthenticationServiceProvider.shared.fetchAccounts(apiService: viewModel.context.apiService)
```
DK1DA, line 95: Yeah we have to do something about this. This doesn't feel right. Created [a ticket for this](https://linear.app/mastodon/issue/IOS-216/fix-fetchaccounts-hack).

Mastodon/Scene/Discovery/ForYou/ProfileCardView.swift
zeitschlag: I checked with the API and decided to make several properties non-optional

Mastodon/Protocol/Provider/DataSourceProvider+NotificationTableViewCellDelegate.swift
zeitschlag: Yes, you're right. In some projects, `case` gets indented, in others it's not. I forgot to toggle the switch 🙈

Mastodon/Protocol/Provider/DataSourceFacade+Status.swift
zeitschlag: Is your question why this method (and the completion) is generic?

Not all menu-actions return a `Mastodon.Entity.Relationship` here. Using the menu you can block a user (returns `Mastodon.Entity.Relationship`), but also a domain returns (`Mastodon.Entity.Empty`) while others like `editStatus` or `translateStatus` do something entirely different.

This was just a first attempt to update a `relationship` IIRC. I consider this WIP and I'm open to other options and ideas!

DK1DA: Ah got you, I thought it might be possible to make `responseToMenuAction` async and to just return `T` instead of having to use a closure. But maybe that's not feasible.

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -41,59 +41,87 @@ extension DataSourceFacade {
...
-        user: ManagedObjectRecord<MastodonUser>
+        username: String,
+        domain: String
     ) async {
-        guard let user = user.object(in: provider.context.managedObjectContext) else {
-            assertionFailure()
-            return
+        provider.coordinator.showLoading()
+
+        Task {
```
DK1DA: This function is already `async` you shouldn't need this additional `Task` here.

Mastodon/Protocol/Provider/DataSourceFacade+Profile.swift
```diff
@@ -41,59 +41,87 @@ extension DataSourceFacade {
...
-        account: Mastodon.Entity.Account
+        domain: String,
+        accountID: String
     ) async {
         provider.coordinator.showLoading()
-        
-        guard let domain = account.domain else { return provider.coordinator.hideLoading() }
-        
+
         Task {
```
DK1DA: I think this `Task` can also be removed.

Mastodon/Scene/Onboarding/Login/MastodonLoginViewController.swift
```diff
@@ -139,30 +139,23 @@ class MastodonLoginViewController: UIViewController, NeedsDependency {
...
-                case .success(let isActived):
-                    assert(isActived)
-                    self.coordinator.setup()
+            .authenticated.sink { (domain, account) in
+                Task {
+                    do {
+                        _ = try await self.context.authenticationService.activeMastodonUser(domain: domain, userID: account.id)
+                        FileManager.default.store(account: account, forUserID: MastodonUserIdentifier(domain: domain, userID: account.id))
+                        Task { @MainActor in
+                            self.coordinator.setup()
```
DK1DA: This feels a bit weird, nesting a `Task` inside a Combine sink, then spawning yet another `Task`.
I actually really like the `asyncMap` approach, and maybe we could annotate `coordinator.setup()` with `@MainActor` so we can get rid of the second `Task` her?

Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
```diff
@@ -238,20 +235,14 @@ final class ProfileHeaderView: UIView {
...
-    
-    required init?(coder: NSCoder) {
-        super.init(coder: coder)
-        _init()
-    }
-    
-}
+    init(account: Mastodon.Entity.Account, me: Mastodon.Entity.Account, relationship: Mastodon.Entity.Relationship?) {
+
+        viewModel = ViewModel(account: account, me: me, relationship: relationship)
```
DK1DA, line 239: I would probably construct the `ProfileHeaderView` with the ViewModel here to only have the three params in the ViewModel's constructor.

Mastodon/Scene/Profile/ProfileViewController.swift
```diff
@@ -202,33 +199,38 @@ extension ProfileViewController {
...
-        .sink { [weak self] isSuspended, isTitleViewDisplaying, tuple1, tuple2 in
-            guard let self = self else { return }
+        .sink { [weak self] account, isTitleViewDisplaying, tuple1, tuple2 in
+            guard let self else { return }
             let (isEditing, _) = tuple1
             let (isMeBarButtonItemsHidden, isReplyBarButtonItemHidden, isMoreMenuBarButtonItemHidden) = tuple2
 
             var items: [UIBarButtonItem] = []
             defer {
-                self.navigationItem.rightBarButtonItems = !items.isEmpty ? items : nil
```
DK1DA, line 219: 🥲 why not just

```
self.navigationItem.rightBarButtonItems = items.isEmpty ? nil : items
```

Isn't this short and concise enough, and maybe even get rid of the `items.isNotEmpty` extension.

Mastodon/Scene/Profile/ProfileViewController.swift
```diff
@@ -202,33 +199,38 @@ extension ProfileViewController {
...
+                if items.isNotEmpty {
+                    self.navigationItem.rightBarButtonItems = items
+                } else {
+                    self.navigationItem.rightBarButtonItems = nil
+                }
             }
 
-            guard !isSuspended else {
+            if let suspended = self.viewModel.account.suspended, suspended == true {
```
DK1DA, line 226: Maybe

```suggestion
            guard self.viewModel.account.suspended == true else {
```

to have it uniform with the other `guard`?

Mastodon/Scene/Profile/ProfileViewController.swift
```diff
@@ -787,101 +754,128 @@ extension ProfileViewController: ProfileHeaderViewControllerDelegate {
...
+
+                self.viewModel.relationship = newRelationship
+                // update account?
+                // update me?
+            }
+
         }
-        
+
+        //            switch relationshipAction {
```
DK1DA: Just a reminder to remove this code.

MastodonSDK/Sources/MastodonCore/Persistence/FileManager+Shared.swift
```diff
@@ -0,0 +1,18 @@
...
+    var documentsDirectory: URL? {
+        urls(for: .documentDirectory, in: .userDomainMask).first
+    }
+
+    var cachesDirectory: URL? {
+        urls(for: .cachesDirectory, in: .userDomainMask).first
+    }
+
+    var sharedDirectory: URL? {
+        containerURL(forSecurityApplicationGroupIdentifier: AppName.groupID)
```
DK1DA, line 16: 👍

MastodonSDK/Sources/MastodonCore/Persistence/FileManager+Timeline.swift
```diff
@@ -1,22 +1,21 @@
 // Copyright © 2023 Mastodon gGmbH. All rights reserved.
 
 import Foundation
-import MastodonCore
 import MastodonSDK
 
 extension FileManager {
```
DK1DA: You could also just make the extension public, then all function will implicitly inherit the access modifier.

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
```diff
@@ -36,55 +35,85 @@ extension Mastodon.Entity {
...
-            case followingCount = "following_count"
-            case moved
-            
-            case fields
-            case bot
-            case source
-            case suspended
-            case muteExpiresAt = "mute_expires_at"
+
+        internal init(id: Mastodon.Entity.Account.ID, username: String, acct: String, url: String, displayName: String, note: String, avatar: String, avatarStatic: String? = nil, header: String, headerStatic: String? = nil, locked: Bool, emojis: [Mastodon.Entity.Emoji] = [], discoverable: Bool? = nil, createdAt: Date, lastStatusAt: Date? = nil, statusesCount: Int, followersCount: Int, followingCount: Int, moved: Mastodon.Entity.Account? = nil, fields: [Mastodon.Entity.Field]? = nil, bot: Bool? = nil, source: Mastodon.Entity.Source? = nil, suspended: Bool? = nil, muteExpiresAt: Date? = nil) {
```
DK1DA: 😱

zeitschlag: Ah! I needed that in the beginning for a `placeholder`, but forgot to remove it afterwards. Thank you!

Mastodon/Scene/Account/AccountListViewModel.swift
```diff
@@ -103,25 +103,25 @@ extension AccountListViewModel {
         authentication: MastodonAuthentication,
         activeAuthentication: MastodonAuthentication
     ) {
-        guard let user = authentication.user(in: context) else { return }
-        
+        guard let account = authentication.account() else { return }
```
DK1DA, line 106: For me this method returns `nil` thus bailing the user out here and leaving the app in a dangling (everything is blank) state.

Steps to reproduce (at least for me):
* Install `develop` branch version
* Log in (with multiple accounts in my case)
* Install this PR branch

Mastodon/Scene/Profile/Header/View/ProfileHeaderView.swift
zeitschlag: I keep thinking: Why not just using the `ProfileViewModel` or use something similar to your `MastodonStatus`?

Mastodon/Scene/Account/AccountListViewModel.swift
zeitschlag: I will look into that, thank you!

Great catch, I can reproduce it, too and I think, I found the culprit: The renaming of `FeedFetchedResultsController` causes this as there was a [`fetchAccounts`](https://github.com/mastodon/mastodon-ios/pull/1192/commits/febbc6f22abc7e62545d16a3e5a29ec463920398#diff-ae17b729d2c09d8dfaffa6380bfabaa33e364f9eb5c0dd48f0a9a655bffdfc47L80-L95) when loading the home-timeline.

I'll look into it how to fix that.

Another way to reproduce this is to remove the account-JSON-files in the shared directory.

zeitschlag: Should work again :)

DK1DA: Yes works now, thank you 👍

Mastodon/Protocol/Provider/DataSourceFacade+Status.swift
zeitschlag: Let me think about it!

Mastodon/Coordinator/SceneCoordinator.swift
```diff
@@ -97,24 +98,53 @@ final public class SceneCoordinator {
...
                                 guard let type = Mastodon.Entity.Notification.NotificationType(rawValue: pushNotification.notificationType) else { return }
                                 guard let authContext = self.authContext else { return }
+                                guard let me = authContext.mastodonAuthenticationBox.authentication.account() else { return }
                                 let notificationID = String(pushNotification.notificationID)
                                 
                                 switch type {
                                     case .follow:
-                                        let profileViewModel = RemoteProfileViewModel(context: appContext, authContext: authContext, notificationID: notificationID)
-                                        _ = self.present(scene: .profile(viewModel: profileViewModel), from: from, transition: .show)
+                                        Task {
```
DK1DA: This feels a bit weird, as we've got a 

```swift
Task {
  DispatchQueue.main.async {
    Task {
      ...
    }
  }
}
```

here, could we maybe eliminate the GCD dispatch and leave it with one task? Maybe some `@MainActor`scoped function used inside etc?

Ref: https://github.com/mastodon/mastodon-ios/pull/1192/files#diff-a97a67b1084ce12242b4ac094ec7976f865dac1a8a18c99323b5fa99d6834981L54

and: https://github.com/mastodon/mastodon-ios/pull/1192/files#diff-a97a67b1084ce12242b4ac094ec7976f865dac1a8a18c99323b5fa99d6834981L78

Mastodon/Protocol/Provider/DataSourceFacade+Block.swift
```diff
@@ -13,56 +13,40 @@ import MastodonSDK
...
-        dependency: NeedsDependency & AuthContextProvider,
-        user: Mastodon.Entity.Account
-    ) async throws {
-        let selectionFeedbackGenerator = await UISelectionFeedbackGenerator()
-        await selectionFeedbackGenerator.selectionChanged()
 
-        let apiService = dependency.context.apiService
-        let authBox = dependency.authContext.mastodonAuthenticationBox
+        let userInfo = [
+            "relationship": response.value,
```
DK1DA: Should we use a string enum here to prevent issues by typos?

Mastodon/Protocol/Provider/DataSourceFacade+Follow.swift
```diff
@@ -7,40 +7,29 @@
...
         let response = try await dependency.context.apiService.toggleFollow(
-            user: user,
+            account: account,
             authenticationBox: dependency.authContext.mastodonAuthenticationBox
         ).value
 
         dependency.context.authenticationService.fetchFollowingAndBlockedAsync()
 
+        NotificationCenter.default.post(name: .relationshipChanged, object: nil, userInfo: [
+            "relationship": response
```
DK1DA: Same question about using a string enum here

Mastodon/Scene/MediaPreview/MediaPreviewViewModel.swift
```diff
@@ -151,8 +151,8 @@ extension MediaPreviewViewModel {
                 return true     // default valid
             case .profileBanner(let item):
                 guard let assertURL = item.assetURL else { return false }
-                guard !assertURL.hasSuffix("missing.png") else { return false }
-                return true
+
+                return assertURL.hasSuffix("missing.png") == false
```
DK1DA: Maybe move this to a string constant somewhere so we can adjust in case the name of this file changes.

Mastodon/Scene/Notification/NotificationView/NotificationView+Configuration.swift
```diff
@@ -0,0 +1,281 @@
...
+            self.acceptFollowRequestActivityIndicatorView.stopAnimating()
+            self.acceptFollowRequestButton.tintColor = .white
+            self.acceptFollowRequestButton.setTitleColor(.white, for: .normal)
+        }
+        if state == .isRejecting {
+            self.rejectFollowRequestActivityIndicatorView.startAnimating()
+            self.rejectFollowRequestButton.tintColor = .clear
+            self.rejectFollowRequestButton.setTitleColor(.clear, for: .normal)
+        } else {
+            self.rejectFollowRequestActivityIndicatorView.stopAnimating()
```
DK1DA, line 269: If `state == .isAccepting` then this code will be executed.
I would use a `switch` (or if/else) here otherwise this code is hard to interpret.

Mastodon/Scene/Profile/Header/ProfileHeaderViewController.swift
```diff
@@ -81,14 +82,30 @@ final class ProfileHeaderViewController: UIViewController, NeedsDependency, Medi
...
+            .store(in: &disposeBag)
+    }
     
+    required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
+
     override func viewDidLoad() {
         super.viewDidLoad()
-        
+
+
```
DK1DA: Excessive blank line here

Mastodon/Scene/Root/Tab.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    var selectedImage: UIImage {
+        return image.withTintColor(Asset.Colors.Brand.blurple.color, renderingMode: .alwaysOriginal)
+    }
+
+    var largeImage: UIImage {
+        return image.withRenderingMode(.alwaysTemplate).resized(size: CGSize(width: 80, height: 80))
+    }
+
+//    @MainActor
```
DK1DA: Reminder: Remove this code.

Mastodon/Scene/Notification/NotificationViewController.swift
```diff
@@ -21,7 +21,7 @@ final class NotificationViewController: TabmanViewController, NeedsDependency {
     var disposeBag = Set<AnyCancellable>()
     var observations = Set<NSKeyValueObservation>()
 
-    var viewModel: NotificationViewModel!
+    var viewModel: NotificationViewModel?
```
DK1DA, line 24: I'm not a huge fan of having those optional view models, because the view itself isn't functional without the VMs. But also there is probably no quick way of changing this, from an architectural perspective.

Mastodon/Scene/Share/View/Content/UserView+Configuration.swift
```diff
@@ -17,68 +17,16 @@ import MastodonSDK
...
         viewModel.relationship = relationship
-        viewModel.user = nil
         self.delegate = delegate
 
         let authorUsername = PlaintextMetaContent(string: "@\(account.username)")
         authorUsernameLabel.configure(content: authorUsername)
 
         do {
-            let emojis = account.emojis?.asDictionary ?? [:]
+            let emojis = account.emojis.asDictionary
```
DK1DA, line 29: ❤️

MastodonSDK/Sources/MastodonCore/AuthenticationServiceProvider.swift
```diff
@@ -101,10 +101,29 @@ public extension AuthenticationServiceProvider {
...
+            guard let account = try? await apiService.accountInfo(domain: authentication.domain,
+                                                                  userID: authentication.userID,
+                                                                  authorization: Mastodon.API.OAuth.Authorization(accessToken: authentication.userAccessToken)).value else { continue }
+
+            FileManager.default.store(account: account, forUserID: authentication.userIdentifier())
+        }
+
+        NotificationCenter.default.post(name: .userFetched, object: nil)
+
+        completion?()
```
DK1DA: Maybe we can get rid of this closure and use the `.userFetched`notification instead?

kimar (approved):
LGTM Code Wise, nothing major, with the exception of [this comment](https://github.com/mastodon/mastodon-ios/pull/1192/files#r1494206691) that might actually lead to issues. Apart from that, good work, going to give it a thorough test now.

Mastodon/Scene/Notification/NotificationViewController.swift
zeitschlag: I'm not a huge fan of forced unwrapping, it feels like a very dirty hack throughout the entire app just to be able to use ViewModels literally everywhere.

But: `?` has a huge advantage over `!`: It doesn't crash, so I consider this a quick workaround.

MastodonSDK/Sources/MastodonCore/AuthenticationServiceProvider.swift
zeitschlag: We can get rid of `completion` as we don't use that anymore. Good catch 👍

Mastodon/Scene/Notification/NotificationView/NotificationView+Configuration.swift
zeitschlag: From what I understand, as I just [copied the code](https://github.com/mastodon/mastodon-ios/commit/7dfa56507e5e2accd7ce8714e748bd159732e7e9#diff-1c984bd601e5b7ddced8d31f5b1b4a98a9e0557b91ec0ee1257b8b78d8a4f745) over:

If `state == .isAccepting`, in fact, if `state` is something else than `isRejecting`, we want the `rejectFollowRequestActivityIndicatorView` to stop spinning (vice versa for the `acceptFollowRequestButton`). Using a `switch` for that would result in lots of code duplication, but maybe I'm wrong?

So yes, there might be more elegant ways, I just didn't want to break anything here :)

DK1DA: Ah, fair enough! Thanks for the clarification, I just re-read the code and it makes sense to me now 👍

Mastodon/Coordinator/SceneCoordinator.swift
zeitschlag: Good catch!

Mastodon/Protocol/Provider/DataSourceFacade+Status.swift
zeitschlag: I thought about it and although this method is already `async`, it can't return anything, as the important stuff happens in `UIAlertAction`-handlers. To my knowledge these handlers don't support return-values. So, if I want something to happen within those handlers, the only way I can think of is a closure.

I also thought about replacing the completion-closure with `.relationshipChanged`-notification. But as the closure already exists and the return-value is used in `DataSourceProvider+NotificationTableViewCellDelegate` to update the menu in `NotificationView`. Sure, we could replace that with a `UIDeferredMenuElement`, but for now I'd prefer to get this PR merged.

kimar (approved):
LGTM 🚀  Great work!

PR #1196, opened 2023-12-15
Use accounts on FavoritedBy/RetootedBy-screens (IOS-214)

zeitschlag:
I came across this when working on the Profile-screen and as this is very isolated, I wanted to extract this. The smaller the PR, the less complexity, the better :)

This PR is about the "Favorited By"/"Retooted By"-screen. So far they used `MastodonUser` and this PR changes them to use `Mastodon.Entity.Account`.

Fixes stuff like this:

IMG_2EAC24C4975F-1


Mastodon/Scene/Profile/UserLIst/UserListViewModel+State.swift
```diff
@@ -123,40 +124,61 @@ extension UserListViewModel.State {
             
             Task {
                 do {
-                    let response: Mastodon.Response.Content<[Mastodon.Entity.Account]>
+                    let accountResponse: Mastodon.Response.Content<[Mastodon.Entity.Account]>
                     switch viewModel.kind {
                     case .favoritedBy(let status):
-                        response = try await viewModel.context.apiService.favoritedBy(
+                            accountResponse = try await viewModel.context.apiService.favoritedBy(
```
DK1DA: This looks like one level of indentation too much, or maybe I'm seeing it wrong?

Mastodon/Scene/Profile/UserLIst/UserListViewModel+State.swift
```diff
@@ -123,40 +124,61 @@ extension UserListViewModel.State {
...
+
                     var hasNewAppend = false
-                    var userIDs = viewModel.userFetchedResultsController.userIDs
-                    for user in response.value {
-                        guard !userIDs.contains(user.id) else { continue }
-                        userIDs.append(user.id)
+
+                    let newRelationships = try await viewModel.context.apiService.relationship(forAccounts: accountResponse.value, authenticationBox: viewModel.authContext.mastodonAuthenticationBox)
+
+                    var accounts = viewModel.accounts
```
DK1DA, line 155: I _think_ this is not doing what you want. You want a copy of the accounts right? This way you'll likely get a reference to the wrapped property.

I've done the same before in the `StatusDataController` (new Name 🎉 ) and had to instantiate an array using the contents of the wrapped property to get a copy: https://github.com/mastodon/mastodon-ios/pull/1167/files#diff-ebefe3e24206519ef132e8d197c0b75529b20348df9de90d6e1465c0f136131eR39

Mastodon/Scene/Profile/UserLIst/UserListViewModel+State.swift
```diff
@@ -123,40 +124,61 @@ extension UserListViewModel.State {
...
+                    for user in accountResponse.value {
+                        guard accounts.contains(user) == false else { continue }
+                        accounts.append(user)
                         hasNewAppend = true
                     }
-                                        
-                    let maxID = response.link?.maxID
-                    
+
+                    var relationships = viewModel.relationships
```
DK1DA, line 163: Same as my previous comments regarding the ref vs. value copy

Mastodon/Scene/Profile/UserLIst/UserListViewModel.swift
```diff
@@ -43,12 +45,8 @@ final class UserListViewModel {
         self.context = context
         self.authContext = authContext
         self.kind = kind
-        self.userFetchedResultsController = UserFetchedResultsController(
```
DK1DA, line 46: 👍

kimar (approved):
LGTM 👍  Just one remark which you might want to look at.

Mastodon/Scene/Profile/UserLIst/UserListViewModel+State.swift
zeitschlag: That's actually a very interesting point, so I had to dig into this 🤓 

As far as I'm aware, structs are value types and as `Array` is a struct, `Array` is also a value type. I also checked the [documentation on swift.org](https://www.swift.org/documentation/articles/value-and-reference-types.html) and they say:

> In Swift, structures, enumerations, and tuples are all value types.

But what about `accounts` as they're `@Published`? Well:

```
(lldb) po type(of: viewModel.accounts)
Swift.Array<MastodonSDK.Mastodon.Entity.Account>
```

Looks like Apple does some black (but transparent) magic here so that it works.
To be honest, I debugged this with a breakpoint in line 159 first 🫣 

```
// before line 159
(lldb) po viewModel.accounts
0 elements
(lldb) po accounts
0 elements
// after line 159
(lldb) po viewModel.accounts
0 elements
(lldb) po accounts
▿ 1 element
  ▿ 0 : <Account: 0x115006640>
```

So to me it looks like it works as expected. But maybe I am missing something here? On the other hand it just works, also on all the other places we've used this concept so far.

PR #1198, opened 2023-12-18
Add Domain Block (IOS-5)

### LINKED ISSUE #364: Missing block domain dropdown option on profiles

zeitschlag:
- Add a new Menu-entry on Profile-Screen only to block/unblock entire domains
- At the moment it's for `MastodonUser` only as #1192 still needs some time, love and labour (but this PR will affect me and my work at #1192, at least slightly)
- Lots of stuff was already prepared.
- Did someone say screenshots? Screenshots (the last one is there to show that all `botsin.space` are blocked):
blockdomain_screenshots

Will fix #364


Mastodon/Scene/Profile/ProfileViewController.swift
```diff
@@ -829,6 +830,27 @@ extension ProfileViewController: ProfileHeaderViewControllerDelegate {
                 let cancelAction = UIAlertAction(title: L10n.Common.Controls.Actions.cancel, style: .cancel, handler: nil)
                 alertController.addAction(cancelAction)
                 present(alertController, animated: true, completion: nil)
+            case .domainBlocking:
+                    guard let user = viewModel.user else { return }
+                    let domain = user.domainFromAcct
+
+                    let alertController = UIAlertController(
+                        title: "Unblock domain \(domain)",
+                        message: "Really unblock domain \(domain)?!",
```
DK1DA: Here `L10n` isn't used yet.

MastodonSDK/Sources/MastodonUI/ViewModel/RelationshipViewModel.swift
```diff
@@ -263,6 +268,10 @@ extension RelationshipViewModel {
             optionSet.insert(.showReblogs)
         }
 
+        if isDomainBlocking {
+            optionSet.insert(.domainBlocking)
```
DK1DA, line 272: I'm wondering if we, somewhere need to remove this attribute incase a domain gets unblocked? However the App likely won'T notice this anyway properly so maybe this is going away with the `MastodonUser` refactoring anyways.

kimar (approved):
LGTM, just one remark regarding missing `L10n`.

Mastodon/Scene/Profile/ProfileViewController.swift
zeitschlag: Oh, good catch, thank you! 🤦

PR #1200, opened 2023-12-20
Show more metadata for posts

### LINKED ISSUE #1199: Sharing toot via share button to iMessage should add post metadata

- [X] I have searched the issues (both open and closed)

### Current Behavior

Sharing toots via the share button through iMessage prevents them from loading link metadata properly. If a link is copy/pasted to iMessage, it will fetch and render the metadata properly.

I tracked this bug down to a likely spot in `URLActivityItemWithMetadata.swift` but without an Apple Developer license I am unable to test any fixes via the Xcode simulator, as it doesn't support iMessage testing.

The impact of this bug is that when viewing a shared link in iMessage, it's harder to see what you're clicking on. The preview feature is very nice, especially for sharing mastodon links to non-fedi friends.

Rendering via share button:
image

Rendering via copy/paste link
image

### Expected Behavior

- Share toot to iMessage
- View toot in iMessage
- It should render the same way that just copying and pasting the URL to iMessage does, where it knows to go fetch the metadata

### Steps To Reproduce

- Share toot to iMessage (any toot should exhibit this behavior)
- View shared toot in iMessage
- There will not be any metadata (images, title, etc) like there should be if you had just copy/pasted the link directly
- If you copy a link to the toot and paste it in iMessage, you will see the metadata.

### Environment

```markdown
- Device: iPhone 15 Pro
- OS: iOS 17.2
- Version: 2023.16
- Build: 6043
```

### Anything else?

_No response_

zeitschlag:
Fixes #1199.

Sets sharing back to Default-URL-Sharing without custom meta data (aka no account icon) so that sharing a status looks like copy'n'pasting that very status in iMessage directly:


kimar (approved):
I assume this "workaround" with manually setting the metadata was from times when iOS didn't correctly resolve Mastodon URLs itself? Looks better to me this way now 👍

Issue #1049, opened 2023-05-16
Server defined post length not respected


RobATX:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Universeodon currently has message limit set to 1,000 characters. The iOS app shows 500, which is the typical default. App should use whatever length is set for the server. 

### Expected Behavior

Use server defined post length/limit

### Environment

```markdown
- Device:
- OS:
- Version:
- Build:
```

j-f1:
Where in the UI are you seeing this? In the in-app post composer UI or in the share sheet → Mastodon option? We’re supposed to be pulling the limit from the API but clearly that isn’t being wired up correctly. Also, does it go away once you start typing?

RobATX:
In the composer. It shows 500 at the bottom. ![image](https://github.com/mastodon/mastodon-ios/assets/31712535/2dbc907f-afac-46b0-841d-fb3bfcf5f80d)

vmstan:
Seeing this now on vmst.io after we raised our limit to 512. Still shows 500 in the app.

vmstan:
This appears to be related to a change introduced in v1.4.7

https://github.com/mastodon/mastodon-ios/blob/1.4.7/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift

```
   // limit
    @Published public var maxTextInputLimit = 500
```

vmstan:
I have confirmed by adjusting this value you can control the post length but I'm not sure how to import the value that server presents in Swift.

100000BC:
I would love to see this get worked on. I am not a coder so I have no idea how to fix it. Does making a new comment here boost the issue? :)

zeitschlag:
Closed in #1207

PR #1207, opened 2024-01-12
#1049 - fix max character limit and other server configs are not respected

kimar:
# Rationale

This fixes server limits which are imposed, e.g. max characters (amongst others) to be respected by the app again.

Apparently, when introducing the v2 instance config, this part of the App has not been ported, thus not respecting the server config anymore as we're not keeping the v1 config if we can query a v2 instance config from the server.

### DIFF
--- a/MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Instance.swift
+++ b/MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Instance.swift
@@ -114,7 +114,7 @@ extension Mastodon.Entity.Instance {
 }
 
 extension Mastodon.Entity.Instance {
-    public struct Configuration: Codable, StatusesMediaAttachmentsPollsContaining {
+    public struct Configuration: Codable, InstanceConfigLimitingPropertyContaining {
         public let statuses: Statuses?
         public let mediaAttachments: MediaAttachments?
         public let polls: Polls?
--- a/MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Instance.swift
+++ b/MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Instance.swift
@@ -114,7 +114,7 @@ extension Mastodon.Entity.Instance {
 }
 
 extension Mastodon.Entity.Instance {
-    public struct Configuration: Codable {
+    public struct Configuration: Codable, StatusesMediaAttachmentsPollsContaining {
         public let statuses: Statuses?
         public let mediaAttachments: MediaAttachments?
         public let polls: Polls?
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
@@ -228,23 +228,33 @@ public final class ComposeContentViewModel: NSObject, ObservableObject {
         }
 
         // set limit
-        if let configuration: Mastodon.Entity.V2.Instance.Configuration = {
-            var configuration: Mastodon.Entity.V2.Instance.Configuration? = nil
-            context.managedObjectContext.performAndWait {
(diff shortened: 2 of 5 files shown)


MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
```diff
@@ -228,34 +228,23 @@ public final class ComposeContentViewModel: NSObject, ObservableObject {
...
         // set limit
-        let _configuration: Mastodon.Entity.Instance.Configuration? = {
+        if let configuration: Mastodon.Entity.V2.Instance.Configuration = {
+            var configuration: Mastodon.Entity.V2.Instance.Configuration? = nil
+            context.managedObjectContext.performAndWait {
+                let authentication = authContext.mastodonAuthenticationBox.authentication
+                configuration = authentication.instance(in: context.managedObjectContext)?.configurationV2
+            }
+            return configuration
+        }() { applyInstanceConfiguration(configuration) }
```
zeitschlag: I really, really dislike this `let ... = { }()`-syntax. So much visual noise.

No need to change it, I just wanted to complain :D

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
```diff
@@ -228,34 +228,23 @@ public final class ComposeContentViewModel: NSObject, ObservableObject {
         }
 
         // set limit
-        let _configuration: Mastodon.Entity.Instance.Configuration? = {
+        if let configuration: Mastodon.Entity.V2.Instance.Configuration = {
+            var configuration: Mastodon.Entity.V2.Instance.Configuration? = nil
+            context.managedObjectContext.performAndWait {
+                let authentication = authContext.mastodonAuthenticationBox.authentication
+                configuration = authentication.instance(in: context.managedObjectContext)?.configurationV2
```
zeitschlag: Memo to self: We could look into storing the `configuration`-JSON on disk to avoid the `context.managedObjectContext.performAndWait`-dance

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Instance.swift
```diff
@@ -114,7 +114,7 @@ extension Mastodon.Entity.Instance {
 }
 
 extension Mastodon.Entity.Instance {
-    public struct Configuration: Codable {
+    public struct Configuration: Codable, StatusesMediaAttachmentsPollsContaining {
```
zeitschlag: What a protocol-name oO Only `Mastodon.Entity.[V2].Instance.Configuration` implement it, right?
What do you think about `SomethingSomethingInstanceConfiguration`?

zeitschlag (approved):
LGTM from a code-perspective 👍 

Feel free to just resolve my comments without changing anything 🙂

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
DK1DA: It would be so cool if Swift supported the Kotlin-esque `?.let {}` here because I think this would be cool and you'd like it, hm now that I'm writing this I'Ve got an idea, let me quickly check

MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
DK1DA: Yes definitely, this will be much appreciated, I'll create a ticket for this.

MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Instance.swift
DK1DA: Yeah I'm also torn about this one 😄  Let me quickly think about it.

PR #1208, opened 2024-01-15
fix: Load missing posts

kimar:
# Rationale

Fixes a bug where the "Load missing posts" button is not showing up anymore / no way to load a gap in the home timeline.


kimar:
This needs more work, will close the PR for now.

PR #1213, opened 2024-01-24
Implement async context menu for statuses

kimar:
# Rationale

To reduce fetching user relationship we could fetch the relationship per-status only wehen the user taps the context menu, as it's only required there.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
+++ b/MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
@@ -717,9 +717,10 @@ extension StatusView.ViewModel {
                                     
                                     menuElement(menu.children)
                                 }
+                            } else {
+                                menuElement(MastodonMenu.setupMenu(actions: [[.shareStatus]], delegate: statusView).children)
                             }
                         }
-                        
                     })
                 ])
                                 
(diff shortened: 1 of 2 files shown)


MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
```diff
@@ -689,21 +685,44 @@ extension StatusView.ViewModel {
...
+                        let domain = authContext.mastodonAuthenticationBox.domain
+
+                        Task { @MainActor in
+                            if let relationship = try? await Mastodon.API.Account.relationships(
+                                session: .shared,
+                                domain: domain,
+                                query: .init(ids: [authorId]),
+                                authorization: authContext.mastodonAuthenticationBox.userAuthorization
+                            ).singleOutput().value {
+                                guard let rel = relationship.first else { return }
```
zeitschlag, line 700: Showing the menu only when the relationship was successfully downloaded implies that a user can't share a status without a connection. That's fine with me.

Also, although the default `timeoutInterval` is 60 seconds, the loading-indicator stays visible ... forever, it seems? oO

zeitschlag (approved):
LGTM 🚀 I added one comment for what feels like missing error handling?

TIL: `UIDeferredMenuElement`, thank you for that!

MastodonSDK/Sources/MastodonUI/View/Content/StatusView+ViewModel.swift
DK1DA: True, I've added a fallback for only share, in case the request fails or times out: https://github.com/mastodon/mastodon-ios/pull/1213/files#diff-a5af8f364c73fb6a18c8c31d7f403f9b15898cd7034f69dd8a775fa9b5461e31R721

PR #1229, opened 2024-02-07
Load User-lists initially (#1227)

zeitschlag:
The `UserList`-screens (Reblogged by/Favorited by) relied on the `ListBatchViewModel` to load their data initially. The `ListBatchViewModel` triggered an update every second. When we changed the time interval for that `ListBatchViewModel` from one to 30 seconds, the screens waited for those 30 seconds to load data.

This PR adds an initial load for those screens, so that it doesn't need to wait for the `ListBatchViewModel` to trigger an update.


kimar (approved):
Ooops :eye: Thx for fixing this!

PR #1230, opened 2024-02-09
#1229, but for status-reporting

zeitschlag:
Trigger status-loading immediately (and not 30 seconds later), see #1227/#1229

(Randomly found during testing)


kimar (approved):
👍

PR #1245, opened 2024-02-29
Fix Profile Editing

### LINKED ISSUE #1244: Can’t edit profile
## Steps to Reproduce the Problem

1. Go to your own profile
2. Tap "Edit Info"

## Specifications

* Version:
* Platform:

zeitschlag:
Probably introduced somewhere around 76304e5 and we just missed that, I guess?

This PR addresses several issues, from my perspective:

- Visibilty of the relationship-action-button
- Editability of the profile (aka setting some `bool` here and there using Combine. It's possible that I missed a spot but for me it works now #famouslastwords)

Fixes #1244, also known as IOS-239.


kimar (approved):
LGTM 👍

PR #1251, opened 2024-03-06
Adjust Server Rules Screen (IOS-220)

kimar:
Updates Server Rules UI

| iPhone | iPad |
|---|---|
| Simulator Screenshot - iPhone 15 Pro Max - 2024-03-06 at 15 35 08|Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2024-03-06 at 15 35 41|


samhenrigold:
Noticing a few things:
- The full-width buttons should be in a platter, anchored to the bottom (`.safeAreaInset(edge: .bottom) { ... }`)
- Background of the platter should use `UIBlurEffect.Style.systemChromeMaterial`
- It should have a thin, 0.5pt top shadow (think how nav bars get a thin separator/shadow thing when you scroll)
  - Shadow color would ideally be `_systemChromeShadowColor`, but that’s private. In lieu of that...
  - Light mode: black at 0.3 alpha
  - Dark mode: white at 0.15 alpha
- Buttons
  - Honestly you could probably rip out most of the custom styling, I’m happy with what the system gives you with a primitive button style + controlSize set to large. Plus it gives us touchDown states for free:
  Screenshot 2024-03-06 at 6 37 57 PM


- Make sure the large title, subtitle (“By continuing...”), and the big number bullets are all aligned
  - I don’t really mind how this happens, either insetting the content to match the large title or nudging the large title outwards to match the content below (if it’s the title provided by the nav bar that might be tricky...)

Beyond those nitpicks, lgtm

kimar:
@samhenrigold makes total sense to me, thanks for the super detailed feedback.
Should we maybe inline the title? It's the system provided large title and looks quite mis-aligned unfortunately :-(

Mastodon/Scene/Onboarding/ServerRules/MastodonServerRulesViewController.swift
```diff
@@ -13,22 +13,97 @@ import MetaTextKit
 import MastodonAsset
 import MastodonCore
 import MastodonLocalization
+import SwiftUI
 
-final class MastodonServerRulesViewController: UIViewController, NeedsDependency {
+struct MastodonServerRulesView: View {
```
zeitschlag, line 18: Why is a SwiftUI-rewrite necessary? What was the problem with the `UITableView`-implementation in combination with a `UINavigationController`?

Mastodon/Scene/Onboarding/ServerRules/MastodonServerRulesViewController.swift
```diff
@@ -38,34 +113,18 @@ extension MastodonServerRulesViewController {
...
 }
 
 extension MastodonServerRulesViewController {
-    @objc private func backButtonPressed(_ sender: UIButton) {
+    @objc private func backButtonPressed(_ sender: UIButton?) {
         navigationController?.popViewController(animated: true)
     }
 
-    @objc private func nextButtonPressed(_ sender: UIButton) {
+    @objc private func nextButtonPressed(_ sender: UIButton?) {
```
zeitschlag, line 151: Why is this button optional now?

zeitschlag (commented):
I wasn't aware that the Onboarding changed again? Maybe I didn't get the memo, but according to Figma, the design hasn't changed, it should look like this:

Screenshot 2024-03-10 at 16 44 15

According to your screenshots, it doesn't. Also I don't get why the SwiftUI-rewrite was necessary? Was the Rules-screen _so broken_, that it had to be rewritten? I just don't understand, maybe you can fill in the gaps? Thank you.

Also there's a color mismatch in your screenshot: The tint color of the back-button in the navigation bar differs from the numbers-color.

Mastodon/Scene/Onboarding/ServerRules/MastodonServerRulesViewController.swift
DK1DA: I don't see why this should be a UITableView and adopting the changes seemed much more straight forward using SwiftUI to me. I wouldn't want to write any new UIKit code if we can use SwiftUI.

Mastodon/Scene/Onboarding/ServerRules/MastodonServerRulesViewController.swift
DK1DA: The selector's target _is_ always an optional value as the target might have been deallocated once this code is reached. Previously we have implicitly unwrapped it. However good point, we can completely remove the target attribute as we've never used it anyways.

Ref: https://developer.apple.com/documentation/uikit/uiapplication/1622946-sendaction

Mastodon/Scene/Onboarding/ServerRules/MastodonServerRulesViewController.swift
zeitschlag: I see your point, but I strongly disagree with this:

> I wouldn't want to write any new UIKit code if we can use SwiftUI.

zeitschlag (approved):
As I wrote elsewhere:

> I won't fight the decision to use SwiftUI from now on

So this PR is fine with me. What just came to my mind is that Onboarding isn't the only place we show the rules, but also in Settings/Server Details:

Simulator Screenshot - iPhone 15 Pro - 2024-03-14 at 12 00 57

But this is probably another issue.

kimar:
@zeitschlag I've updated the server details to use the new view:

Simulator Screenshot - iPhone 15 - 2024-03-14 at 16 11 18

PR #1256, opened 2024-03-12
Don't add my own handle to a reply (#1254)

### LINKED ISSUE #1254: My own handle is added to the reply
> Another bug: when replying to someone that mentioned me, my own handle is added to the reply

zeitschlag:
Fixes #1254. This was a nice little bug. This caused it:

```swift
let mentions = status.mentions ?? []
    .filter { author?.id != $0.id }
// ...
```

which can be read as 

```swift
let mentions = status.mentions ?? [].filter { author?.id != $0.id }
```

So, in case there're no mentions, we filter an empty list to not include our own handle.
The fix makes the optional `mentions` required, as [according to the docs](https://docs.joinmastodon.org/entities/Status/#mentions), it's not even optional.

On another note: We should look into what it takes to remove `Status`-persistence.


kimar (approved):
👍

PR #1259, opened 2024-03-19
Fix Profile Editing again (IOS-239, #1244)

### LINKED ISSUE #1244: Can’t edit profile

zeitschlag:
This needs a bit of explanation, I guess, so please don't squash if possible? I didn't take into consideration, that the `ProfileViewController.viewModel` changes. And when it changes, all the combine-connections just ... disappear. This PR changes that (but probably I oversaw something again).


kimar:
@zeitschlag thank you for the fix! I'll take a thorough look 👀

kimar (approved):
Looks good to me, the only thing I noticed is that, if there are no changes to the mode, upon pressing "Save info" nothing happens. Maybe this should actually have the same effect as pressing cancel?

zeitschlag:
@kimar good suggestion, thank you!

PR #1261, opened 2024-03-19
Explore Tab: Use a segmented picker under the search bar (IOS-237)

zeitschlag:
- Replace the Tab-menus in Explore with a UISegmentedControl
- First idea was to use `UISearchBar` with its `scopeTitles` and this also worked in iPhone, but not on iPad (for whatever reason the navigation bar height just doesn't change)
- So I got back to use a normal `UISegmentedControl` (but not as part of the navigation bar) and it looks like this:

Simulator Screenshot - iPhone 15 Pro - 2024-03-22 at 17 26 14
Simulator Screenshot - iPad (10th generation) - 2024-03-22 at 17 20 19

LMK if you want me to add a thin line below the segmented control

### DIFF
--- a/Mastodon/Scene/Discovery/DiscoveryViewController.swift
+++ b/Mastodon/Scene/Discovery/DiscoveryViewController.swift
@@ -12,8 +12,6 @@ import MastodonAsset
 import MastodonCore
 import MastodonUI
 
-
-
 public class DiscoveryViewController: PageboyViewController, NeedsDependency {
 
     public static let containerViewMarginForRegularHorizontalSizeClass: CGFloat = 64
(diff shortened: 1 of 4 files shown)


Mastodon/Scene/Discovery/DiscoveryViewController.swift
```diff
@@ -7,14 +7,15 @@
...
 import Combine
-import Tabman
 import Pageboy
 import MastodonAsset
 import MastodonCore
 import MastodonUI
 
-public class DiscoveryViewController: TabmanViewController, NeedsDependency {
-    
+
```
DK1DA: Did you intend to add those blank lines?

kimar (approved):
LGTM 👍

PR #1267, opened 2024-04-05
Move all UI*FeedbackGenerators to FeedbackGenerator and disable them for now (IOS-247)

kimar:


Mastodon/Supporting Files/SceneDelegate.swift
```diff
@@ -22,10 +22,14 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     var coordinator: SceneCoordinator?
 
     var savedShortCutItem: UIApplicationShortcutItem?
+    
+    let feedbackGenerator = FeedbackGenerator.shared
 
     func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
         guard let windowScene = scene as? UIWindowScene else { return }
         
+        feedbackGenerator.isEnabled = false // Disable Haptic Feedback for now
```
DK1DA, line 31: Disabled by now

MastodonSDK/Sources/MastodonCore/FeedbackGenerator.swift
```diff
@@ -0,0 +1,41 @@
+// Copyright © 2024 Mastodon gGmbH. All rights reserved.
```
zeitschlag, line 1: I saw that you already changed the previous `File.swift`-header 👍

MastodonSDK/Sources/MastodonCore/FeedbackGenerator.swift
```diff
@@ -0,0 +1,41 @@
...
+
+import UIKit
+
+public class FeedbackGenerator {
+    
+    private init(){}
```
zeitschlag: Is this empty `init` necessary? Having it before all the properties also feels weird

zeitschlag (commented):
Nice! 👍

MastodonSDK/Sources/MastodonCore/FeedbackGenerator.swift
DK1DA: I'd like to prevent one from initializing another instance of this class, otherwise the default constructor inherits the visibility from the class.

I mean I could also make this an enum, this would also prevent initialization. What do you think?

zeitschlag: Ah! I didn't get that this is your intention, thank you. From what I know, an enum is _the_ way to go in that case, but maybe that changed? `enum`s don't support properties, do they? So those generators need be static, right? Also not elegant.

I've considered the `shared`-instances/singletons always a convenience-thing, something additional. If people wanted to create an own instance of that class, well, they still would be free to do so. I mean you still could add a little comment `// I don't want you to create your own instance of FeedbackGenerator, use .shared, bloody hell` as subtle hint to your intentions?

zeitschlag (approved):
Sorry, forgot to approve as nothing is blocking :)

MastodonSDK/Sources/MastodonCore/FeedbackGenerator.swift
DK1DA: I'm removing the private constructor, I don't see a strong reason to keep it here. It might make sense to have it in case there is a reason to not allow users to construct their own instance, e.g. because of shared mutable state etc. but that's not the case here.

PR #1272, opened 2024-04-13
Post cell cleanup (IOS-232) and other fixes

zeitschlag:
- Uses `RelativeDateTimeFormatter` with abbreviated units
- Removes the (new SF symbols)-visibility indicator
- Replaces some icons with SF Symbols
- Also: Fixes the TabBar-item-icon of the Home-tab
- Also: Fixes the icon-position in the TimelineSelector-button (I noticed that in the different screenshots)

old (with correct icon-position):

IMG_281C7A7F20D3-1

✨ New ✨ (with not so correct icon-position):

Simulator Screenshot - iPhone 15 Pro - 2024-04-13 at 15 25 01


kimar (approved):
LGTM 👍

PR #1278, opened 2024-04-22
Fix 💊-Issues

### LINKED ISSUE #1275: 💊 -Issues
- New Posts-Pill is visible when switching between Local and Following-Timelines
- Doesn't disappear when scrolling to top

zeitschlag:
- When switching the timelines, the pill appeared with "New Posts!!!!!!" which makes sense but also it doesn't.
- When scrolling to the top, the pill didn't disappear. This annoyed me so I made it go away when timeline was scrolled to top.

Fixes #1275.


kimar (approved):
LGTM 👍

PR #1279, opened 2024-04-23
Fix array oob crash (IOS-257)

kimar:
Supposed to fix an out-of-bounds crash:

Bildschirmfoto 2024-04-23 um 14 26 12

### DIFF
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
@@ -105,9 +105,10 @@ extension HomeTimelineViewModel.LoadLatestState {
 
         guard let viewModel else { return }
 
-        let latestFeedRecords = viewModel.dataController.records.prefix(APIService.onceRequestStatusMaxCount)
 
-        Task {
+        Task { @MainActor in
+            let latestFeedRecords = viewModel.dataController.records.prefix(APIService.onceRequestStatusMaxCount)
+
             let latestStatusIDs: [Status.ID] = latestFeedRecords.compactMap { record in
                 return record.status?.reblog?.id ?? record.status?.id
             }
@@ -128,7 +129,7 @@ extension HomeTimelineViewModel.LoadLatestState {
                     )
                 }
 
-                await enter(state: Idle.self)
+                enter(state: Idle.self)
                 viewModel.receiveLoadingStateCompletion(.finished)
 
                 // stop refresher if no new statuses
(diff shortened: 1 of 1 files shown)


Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -105,9 +105,10 @@ extension HomeTimelineViewModel.LoadLatestState {
 
         guard let viewModel else { return }
 
-        let latestFeedRecords = viewModel.dataController.records.prefix(APIService.onceRequestStatusMaxCount)
 
-        Task {
+        Task { @MainActor in
+            let latestFeedRecords = viewModel.dataController.records.prefix(APIService.onceRequestStatusMaxCount)
```
zeitschlag, line 110: I don't think this has something to do with the index-fix, but still it crossed my mind:

Do you know or remember the reason for `prefix(100)`? A few lines below we use the IDs of `latestFeedRecords` to determine new statuses: A status is new, if its ID isn't among those of max. 100 `latestFeedRecords` (from my understanding). But what if its the 101. or 102.?

I didn't find anything in the git history.

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -146,7 +147,7 @@ extension HomeTimelineViewModel.LoadLatestState {
                         for (i, record) in newRecords.enumerated() {
                             if let index = oldRecords.firstIndex(where: { $0.status?.reblog?.id == record.id || $0.status?.id == record.id }) {
                                 oldRecords[index] = record
-                                if newRecords.count > index {
+                                if newRecords.count > i {
```
zeitschlag, line 150: Great catch!

zeitschlag (approved):
As I said: I could reproduce this bug when I lowered the `onceRequestStatusMaxCount` to `10` and this PR (`i` instead of `index`) fixed it. Thank you for that!

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
zeitschlag: I'm asking as using all the records (and not only the `n` most recent onces) seems to fix the weird timeline-sorting issue I mentioned ... somewhere (although some posts are missing then in between. This is super strange)

DK1DA: Maybe the hard-coded value is related to the max result count of the (paginated) API? There should be a "Load missing posts" divider if there is a gap in between.

zeitschlag: Yes, I've seen that cell before, but it doesn't show up.

`limit` in the API-docs defaults to 20 and has a maximum value of 40. So using 100 as clientside default doesn't make a lot of sense to me.

I'll look into it :)

PR #1281, opened 2024-04-29
Scrolling crash content offset (IOS-254)

zeitschlag:
- This PR is an attempt to fix our most prominent crash. As I haven't been able to reproduce it, I can't say for sure that it's fixed. We'll find out.
- It fixes another issue introduces with #1275: The pill disappeared immediately
- It also makes sure to only persist the `following`-timeline (but not the `public`-timeline).


kimar (approved):
LGTM 👍

PR #1283, opened 2024-05-03
Fix "Load More" Button on Home/Public Timeline

kimar:
This PR fixes the broken Gap Loading aka. "Load More".

Currently the "Load More"-button is never shown on the timeline, but there might be gaps in the timeline.

### DIFF
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
@@ -140,8 +140,9 @@ extension HomeTimelineViewModel.LoadLatestState {
                     viewModel.didLoadLatest.send()
                 } else {                    
                     viewModel.dataController.records = {
-                        var newRecords: [MastodonFeed] = newStatuses.map {
-                            MastodonFeed.fromStatus(.fromEntity($0), kind: .home)
+                        var newRecords: [MastodonFeed] = newStatuses.map { status in
+                            let hasMore = status == newStatuses.last
+                            return MastodonFeed.fromStatus(.fromEntity(status), kind: .home, hasMore: hasMore)
                         }
                         var oldRecords = viewModel.dataController.records
                         for (i, record) in newRecords.enumerated() {
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
@@ -116,15 +116,20 @@ extension HomeTimelineViewModel.LoadLatestState {
             do {
                 await AuthenticationServiceProvider.shared.fetchAccounts(apiService: viewModel.context.apiService)
                 let response: Mastodon.Response.Content<[Mastodon.Entity.Status]>
-
+                
+                /// To find out wether or not we need to show the "Load More" button
+                /// we have make sure to eventually overlap with the most recent cached item
+                let sinceID = latestFeedRecords.count > 1 ? latestFeedRecords[1].id : "1"
+                
                 switch viewModel.timelineContext {
                 case .home:
                     response = try await viewModel.context.apiService.homeTimeline(
+                        sinceID: sinceID,
                         authenticationBox: viewModel.authContext.mastodonAuthenticationBox
                     )
                 case .public:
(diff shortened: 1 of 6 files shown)


Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -116,15 +116,20 @@ extension HomeTimelineViewModel.LoadLatestState {
             do {
                 await AuthenticationServiceProvider.shared.fetchAccounts(apiService: viewModel.context.apiService)
                 let response: Mastodon.Response.Content<[Mastodon.Entity.Status]>
-
+                
+                /// To find out wether or not we need to show the "Load More" button
+                /// we have make sure to eventually overlap with the most recent cached item
+                let sinceID = latestFeedRecords.count > 1 ? latestFeedRecords[1].id : "1"
```
zeitschlag, line 122: What's the reascon `sinceID` is not an optional?
Both Timeline-calls support `sinceID` being `nil` and `1` as fallback confused me

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
```diff
@@ -146,32 +146,67 @@ extension HomeTimelineViewModel {
...
+                hasMore = false
+            } else {
+                /// if fetched items and first item after gap don't match -> we got another gap
+                hasMore = item != head.first?.status?.entity
+            }
+
+            feedItems.append(
+                .fromStatus(item.asMastodonStatus, kind: .home, hasMore: hasMore)
+            )
+        }
```
zeitschlag, line 204: This looks pretty similar to the code above in `ViewModel+LoadLatestState`, right? What do you think about a little comment that links those two?

It's about pointing out that there's a connection between those two, even though there're not identical :)

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
```diff
@@ -146,32 +146,67 @@ extension HomeTimelineViewModel {
...
         // fetch data
-        let maxID = status.id
-        _ = try? await context.apiService.homeTimeline(
-            maxID: maxID,
-            authenticationBox: authContext.mastodonAuthenticationBox
-        )
+        let response: Mastodon.Response.Content<[Mastodon.Entity.Status]>?
         
-        record.isLoadingMore = false
+        switch timelineContext {
```
zeitschlag, line 161: Good catch! I admit: I didn't see this when adding the `timelineContext` 🫠

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
```diff
@@ -146,32 +146,67 @@ extension HomeTimelineViewModel {
 extension HomeTimelineViewModel {
 
     // load timeline gap
-    func loadMore(item: StatusItem) async {
+    @MainActor
```
zeitschlag, line 149: What's th reason to run this entire code on `Main`, including the download?

zeitschlag (approved):
LGTM, works like a charm! Thank you for your work. 👍 🚀 

As already mentioned elsewhere: The order of the statuses is weird sometimes after I pressed the "Load More"-button. I saw that using the displayed timestamps (I didn't run a debugger or so, sorry). What I could imagine is that it's about `maxID` and `sinceID`? But honestly: I don't know and don't consider this something blocking as it looks more like a UI-issue to me.

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
DK1DA: I have adopted this from the Android implementation, there `sinceID` is also optional but when loading the Gaps it is explicitly using `1` which makes me assume the API has a different behavior when not setting the value at all.

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
DK1DA: The download is run off the mainThread because `dataTaskPublisher` will dispatch to a worker thread by default. Only the code inside this function is run explicitly on the main thread to ensure data consistency when modifying the status records.

Bildschirmfoto 2024-05-08 um 09 18 08

Issue #1206, opened 2024-01-08
Refactor ReportStatusTableViewCell


kimar:
# Rationale

The `ReportStatusTableViewCell` only contains a very shallow ViewModel which could be removed with it's contained properties being added to the Cell directly.

somkuwaryash:
Hey,can i take on this issue?

kimar:
@somkuwaryash sure 👍  please go ahead and raise a PR once you're ready with the changes.

zeitschlag:
I was about to write that, but Marcus was faster :D

@somkuwaryash Thank you for your offer, we really appreciate that!
If there's anything you need help with: Please ask, we're here to help :)

ali-ahsan-ali:
Seeing as it has been 5 months. Can I also take a look at this issue. Could be a race to the solution ;)

ali-ahsan-ali:
Race is over? That was quick. Maybe I should have left this for someone newer to programming in swift. Atleast I grew some confidence and maybe can tackle some harder issues. 

Please take a look @zeitschlag @kimar

PR #1285, opened 2024-05-06
Issue-1206 - Remove unnecessary view model

ali-ahsan-ali:
https://github.com/mastodon/mastodon-ios/issues/1206

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -217,6 +217,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  arm64-darwin-23
   x86_64-darwin-21
   x86_64-darwin-22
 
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -217,6 +217,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  arm64-darwin-23
```
ali-ahsan-ali, line 220: I acknowledge this change. I would like an explanation of why NOT this should be added to the gem file. I think from research this specifies the supported platforms. Seeing as I just built the project on said platform, I reckon it is supported. Let me know and I can rip this one out, but want to take this opportunity of contribution to also learn something if possible.

PR #1287, opened 2024-05-07
Moderation Warning Notifications (IOS-264)

zeitschlag:
In case you've been a naughty person and you received a warning from moderation, you will see that as a notification (from now on). There are different types of warnings (see [here](https://github.com/mastodon/mastodon/pull/30065) for details). Users can see details for those warnings when they tap an the cell, they get redirected to the website.

This PR also re-adds the FILE_HEADER-macros and adds authentication to instance-calls.

On iOS it looks like this:

Simulator Screenshot - iPhone 15 Pro - 2024-05-09 at 22 40 22


Mastodon/Scene/Onboarding/PickServer/MastodonPickServerViewModel.swift
```diff
@@ -167,7 +167,7 @@ extension MastodonPickServerViewModel {
                 self.unindexedServers.value = nil
                 return self.context.apiService.webFinger(domain: domain)
                     .flatMap { domain -> AnyPublisher<Result<Mastodon.Response.Content<[Mastodon.Entity.Server]>, Error>, Never> in
-                        return self.context.apiService.instance(domain: domain)
+                        return self.context.apiService.instance(domain: domain, authenticationBox: nil)
```
DK1DA, line 170: @zeitschlag as per API documentation none of the endpoint we're using for `instance` or `v2/instance` requires authentication, could you please explain why this is required here?

zeitschlag: Sure! As I learned the other day, when you can't just call a `LIMITED_FEDERATION_INSTANCES`, you need authentication for these endpoints (for whatever reason). This is also explained in the commit-message for this change, so it might be good, _not_ to squash this time :)

Someone should fix the API-documentation at some point, probably? @renchap 👀

DK1DA: Well that makes sense then! Yes would probably be good to have this documented somewhere.

kimar (approved):
👍

PR #1299, opened 2024-05-21
Don't flicker avatar-image in MainTabBar (IOS-238)

zeitschlag:
Every time, the account gets updated, the `avatarURL` is updated as well. But not only new image was set. Before setting the image, the button-image was reset to the placeholder first. After that, the downloaded image was applied.

Now we don't reset the image to the placeholder, but set the placeholder once.
Also: Cleanup.


kimar (approved):
Ah, great work! 👍  Much better now :-) 

I noticed that, when I don't have an avatar, the default one still has a _very_ minimal flickering when returning from a user profile. But I don't consider this a blocker in any way.

PR #1303, opened 2024-05-23
Improve "Load more"-times when scrolling to the end (IOS-272)

zeitschlag:
This is basically a refactoring of the `ListBatchFetchViewModel`, it does two things:

1. Remove the "Reload this every 30 seconds", which caused a delay on several screens
2. When users reach the bottom of the ScrollView, new content is requested
3. DIY: Use this mechanism everywhere

The previous mechanism worked like this: Check every second, if the user reached the end of the scrollView. If so: `shouldFetch`. As we increased "every second" to "30 seconds", this caused a significant delay on all screens which used this mechanism.

Others brought their own solution to fetch new content when users reached THE END, like the HomeTimeline or the FollowerList. From now one, there's a suggested way to deal with this "We must do something once the user reachs the end!!!!"-issue.

P.S.: I'm not so happy with the `Self.`-approach and if someone has a better name for `scrollViewDidScrollToEnd`, I'd be also thankful for a hint. Maybe it's me, but I would have loved to add a new method to the `UIScrollViewDelegate`-protocol (without Protocol-inheritance). Maybe someone knows a way to do this?

### DIFF
--- a/Mastodon.xcodeproj/project.pbxproj
+++ b/Mastodon.xcodeproj/project.pbxproj
@@ -365,7 +365,7 @@
 		DB6D9F3526351B7A008423CD /* NotificationService+Decrypt.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB6D9F3426351B7A008423CD /* NotificationService+Decrypt.swift */; };
 		DB72601C25E36A2100235243 /* MastodonServerRulesViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB72601B25E36A2100235243 /* MastodonServerRulesViewController.swift */; };
 		DB72602725E36A6F00235243 /* MastodonServerRulesViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB72602625E36A6F00235243 /* MastodonServerRulesViewModel.swift */; };
-		DB7274F4273BB9B200577D95 /* ListBatchFetchViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB7274F3273BB9B200577D95 /* ListBatchFetchViewModel.swift */; };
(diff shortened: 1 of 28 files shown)


kimar:
@zeitschlag regarding

> I'm not so happy with the Self.-approach and if someone has a better name for scrollViewDidScrollToEnd, I'd be also thankful for a hint. Maybe it's me, but I would have loved to add a new method to the UIScrollViewDelegate-protocol (without Protocol-inheritance). Maybe someone knows a way to do this?

You could create a custom e.g. `TimelineScrollViewDelegate` which conforms to `UIScrollViewDelegate` and provides default implementations as well as your new delegate protocol method. This way you could remove the `Self.` and the `action` closure with the drawback that this adds more glue code.

zeitschlag:
> You could create a custom e.g. `TimelineScrollViewDelegate`

That's an idea, I might have to think about it a little longer as I don't think it's the best solution to implement the `UIScrollViewDelegate` in that class. For example: In `HomeTimelineViewController` we implement some `UIScrollViewDelegate`-methods as we need them to hide the pill. Replacing that implementation with a default-implementation doesn't sound very helpful to me. But it's a good hint, thank you!

kimar:
@zeitschlag 

> That's an idea, I might have to think about it a little longer as I don't think it's the best solution to implement the UIScrollViewDelegate in that class. For example: In HomeTimelineViewController we implement some UIScrollViewDelegate-methods as we need them to hide the pill. Replacing that implementation with a default-implementation doesn't sound very helpful to me. But it's a good hint, thank you!

Yes that's what I was also thinking, you could use the custom delegate to proxy calls to the other class and keep the custom implementations there, but then again it will result in more complexity.

kimar (approved):
LGTM 👍

PR #1306, opened 2024-05-30
IOS-168: Implement Privacy & Safety Settings

kimar:
| Rationale | Demo |
|---|---|
| Implements a settings screen to control your Privacy & Safety settings from within the App. | RocketSim_Recording_iPhone_15_6 1_2024-05-30_11 58 40 |


MastodonSDK/Sources/MastodonSDK/Entity/Mastodon+Entity+Account.swift
```diff
@@ -37,6 +37,8 @@ extension Mastodon.Entity {
         public let locked: Bool
         public let emojis: [Emoji]
         public let discoverable: Bool?
+        public let hideCollections: Bool?
+        public let indexable: Bool?
```
zeitschlag: If I have a look at https://docs.joinmastodon.org/entities/account/, I don't find these values. Where do you get these from?

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
```diff
@@ -0,0 +1,217 @@
...
+            preset = .openPublic
+        } else if self == Self.privateRestricted {
+            preset = .privateRestricted
+        } else {
+            preset = .custom
+        }
+    }
+
+    func loadSettings() {
+        Task { @MainActor in
```
zeitschlag, line 140: Just out of curiosity: Is there a difference between `Task { @MainActor in` and the `@MainActor`-attribute?

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
```diff
@@ -0,0 +1,217 @@
...
+                domain: domain,
+                query: .init(
+                    discoverable: suggestMyAccountToOthers,
+                    locked: manuallyApproveFollowRequests,
+                    source: .withPrivacy(visibility.toPrivacy()),
+                    indexable: appearInSearches,
+                    hideCollections: !showFollowersAndFollowing
+                ),
+                authorization: userAuthorization
+            ).value
```
zeitschlag: What do we need this `.value` for?

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewController.swift
```diff
@@ -0,0 +1,119 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        title = L10n.Scene.Settings.PrivacySafety.title
+    }
+}
+
+struct PrivacySafetyView: View {
```
zeitschlag: I was a bit surprised to see this `View` (and the `CheckableButton`) in the same file as the `PrivacySafetyViewController`, would it make sense for you to move them into their own files?

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
```diff
@@ -0,0 +1,217 @@
...
+        vm.visibility = .public
+        vm.manuallyApproveFollowRequests = false
+        vm.showFollowersAndFollowing = true
+        vm.suggestMyAccountToOthers = true
+        vm.appearInSearches = true
+        return vm
+    }()
+    
+    static let privateRestricted: PrivacySafetyViewModel = {
+        let vm = PrivacySafetyViewModel(appContext: nil, authContext: nil, coordinator: nil)
```
zeitschlag: I needed a moment to understand, why the parameters are nil, but from my understanding, `openPublic` and `privateRestricted` both serve as templates, right? Their `appContext` etc. is totally irrelevant, as their other values get applied to a `PrivacySafetyViewModel`, which already has a `appContext, ...`, right?

zeitschlag (approved):
Works like a charm 👍 

As always: Some (more or less dumb) questions with one exception: Where do these new properties in `Mastodon.Entity.Account` come from? They don't show up in the docs

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
DK1DA: I wanted to use a `Task` in this place to not have the ` loadSettings` function have to be invoked in an async context. But other than that i think it shouldn't make a difference.

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
DK1DA: Oh I think that's a debug left-over, will remove it.

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewController.swift
DK1DA: Yes I'm doing this 👍

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
DK1DA: Yes these defintions are only used for comparing agains the current settings and applying presets, so the context here is not consumed. Now that I'm writing this, I think the cleaner solution would be to use protocols for this, I'll fix this.

DK1DA: And it's done, thanks for the Feedback!

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
```diff
@@ -0,0 +1,225 @@
...
+        vm.visibility = .public
+        vm.manuallyApproveFollowRequests = false
+        vm.showFollowersAndFollowing = true
+        vm.suggestMyAccountToOthers = true
+        vm.appearInSearches = true
+        return vm
+    }()
+    
+    static let privateRestricted: PrivacySafetyViewModel = {
```
zeitschlag: I wondered why you used the protocol above, but here the ViewModel?

```suggestion
    static let privateRestricted: PrivacySafetySettingApplicable = {
```

zeitschlag (approved):
👍

Mastodon/Scene/Settings/Privacy and Safety/PrivacySafetyViewModel.swift
DK1DA: That was actually a leftover this whole code, except from the `apply(from:)` method, can go . I'll remove it.

PR #1308, opened 2024-06-06
Remove Poll Option and Poll Persistence (IOS-182)

zeitschlag:
Removes persistence for Poll Option (and also Polls, while I was at it)


kimar (approved):
LGTM 👍

PR #1311, opened 2024-06-11
Make the CI clean and build (instead of just clean) (IOS-280)

zeitschlag:


kimar (approved):
🤦  I would have expected for `build: true` to be the default in `xcodebuild`. Good catch!

PR #1316, opened 2024-06-13
Bugfixes for "Load More" / Home Timeline Gap Loading (IOS-275)

### LINKED ISSUE #1288: [BUG] “Load missing posts” appears when there isn’t any

- [X] I have searched the issues (both open and closed)

### Current Behavior

The “load missing posts” timeline item appears when there is no posts between the old and new timelines. So tapping it just makes it disappear and connect the two timeline sections 

### Expected Behavior

For the button to not be present when it’s not necessary 

### Steps To Reproduce

1. Open the app after a short time
2. Wait for the timeline to load and show “load missing posts”
3. Tap “load missing posts”
4. Observe no posts are loaded and the timelines are now connected

### Environment

```markdown
- Device: iPhone 12 Pro
- OS: iOS 17.4.1
- Version: v2024.5
- Build: 6336
```

### Anything else?

https://github.com/mastodon/mastodon-ios/assets/8526392/cb04673c-27f8-45de-9d52-b320df0a8d22

kimar:
Fixes #1288 and incorrectly ordered posts when loading more.

### DIFF
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
@@ -154,7 +154,7 @@ extension HomeTimelineViewModel.LoadLatestState {
                         toAdd = statuses.map({ MastodonFeed.fromStatus($0.asMastodonStatus, kind: .home) })
                         toAdd.last?.hasMore = true
                     } else {
-                        /// We do not have existing items, no load more is required as there can't be a gap
+                        /// We do not have existing items, no _Load More_ is required as there is no gap
                         toAdd = statuses.map({ MastodonFeed.fromStatus($0.asMastodonStatus, kind: .home) })
                     }
                     
--- a/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
+++ b/Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
@@ -125,12 +125,11 @@ extension HomeTimelineViewModel.LoadLatestState {
                 case .home:
                     response = try await viewModel.context.apiService.homeTimeline(
                         sinceID: sinceID,
-                        limit: 20,
                         authenticationBox: viewModel.authContext.mastodonAuthenticationBox
                     )
                 case .public:
                     response = try await viewModel.context.apiService.publicTimeline(
-                        query: .init(local: true, sinceID: sinceID, limit: 20),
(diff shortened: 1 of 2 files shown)


Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
```diff
@@ -163,12 +163,11 @@ extension HomeTimelineViewModel {
         case .home:
             response = try? await context.apiService.homeTimeline(
                maxID: status.id,
-               limit: 20,
                authenticationBox: authContext.mastodonAuthenticationBox
            )
         case .public:
             response = try? await context.apiService.publicTimeline(
-                query: .init(maxID: status.id, limit: 20),
+                query: .init(maxID: status.id),
```
zeitschlag: I guess you removed `limit` as the public timeline [defaults to 20 posts](https://docs.joinmastodon.org/methods/timelines/#public) nevertheless?

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -119,7 +119,7 @@ extension HomeTimelineViewModel.LoadLatestState {
                 
                 /// To find out wether or not we need to show the "Load More" button
                 /// we have make sure to eventually overlap with the most recent cached item
-                let sinceID = latestFeedRecords.count > 1 ? latestFeedRecords[1].id : "1"
+                let sinceID = latestFeedRecords.count > 1 ? latestFeedRecords[1].id : nil
```
zeitschlag, line 122: Thank you!

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
zeitschlag:
```suggestion
                query: .init(maxID: status.id),
                query: .init(local: true, maxID: status.id),
```

Just saw posts from other instances appearing in my local timeline 🫠

zeitschlag (commented):
Most of the stuff looks good, it works most of the time, here it doesn't and I don't really get why:

D579C99E-0C9B-453D-A544-30D8F2CC7241_1_201_a

My guess is that there's a flaw when caclulating the `hasMore` in "load more", my culprit is the `if index < items.count - 1`-condition, but this is just a wild guess.

Maybe you could have another look at that?

zeitschlag (approved):
Still looks good from my perspective, I don't know how to fix the "Scroll to the bottom"-thing I sent you, but it's not a blocker for me, to be honest, as it only seems to appear, when you scroll to the bottom fast.

PR #1319, opened 2024-06-25
Notification Filtering (IOS-241)

zeitschlag:
- Adds a new cell on the "Everything"-notification-screen to show unwanted notifications
- Users can define what "unwanted" means to them using the new "Filter"-button in the upper right corner of the Notification-screen
- Filtered notifications are sorted by account and users can dismiss/accept if they want to get notifications of that user (it's some standard table-views and delegates)

## Screenshots

ios_241_1

ios_241_2


Mastodon/Protocol/Provider/DataSourceFacade+Notifications.swift
```diff
@@ -0,0 +1,52 @@
...
+
+            if provider.traitCollection.userInterfaceIdiom == .phone {
+                transition = .show
+            } else {
+                transition = .modal(animated: true)
+            }
+
+            provider.coordinator.present(scene: .notificationRequests(viewModel: viewModel), transition: transition)
+        } catch {
+            //TODO: Error Handling
```
DK1DA, line 30: @zeitschlag is there anything we should add here for error handling?

Mastodon/Protocol/Provider/DataSourceProvider+NotificationTableViewCellDelegate.swift
```diff
@@ -514,10 +514,8 @@ extension NotificationTableViewCellDelegate where Self: DataSourceProvider & Aut
                 )
             case .account(let account, _):
                 await DataSourceFacade.coordinateToProfileScene(provider: self, account: account)
-            case .notification:
-                assertionFailure("TODO")
-            case .hashtag(_):
-                assertionFailure("TODO")
+            case .notification, .hashtag(_), .notificationBanner(_):
+                print("TODO")
```
DK1DA: This feels a bit weird, I mean is this a case that's only covered to cover all enum cases or can this code really be executed?

Mastodon/Protocol/Provider/DataSourceProvider+StatusTableViewCellDelegate.swift
```diff
@@ -618,9 +618,7 @@ extension StatusTableViewCellDelegate where Self: DataSourceProvider & AuthConte
                     provider: self,
                     account: account
                 )
-            case .notification:
-                assertionFailure("TODO")
-            case .hashtag(_):
+            case .notification, .hashtag(_), .notificationBanner(_):
                 assertionFailure("TODO")
```
DK1DA: Basically same question as here: https://github.com/mastodon/mastodon-ios/pull/1319/files#diff-2bf0f1a0f7802c1cb4943a5515edda158dd617c42be9e2c81465709a1e37499cR518

If this is never invoked maybe a `break` would be sufficient?

Mastodon/Scene/Notification/Cell/NotificationFilteringBannerTableViewCell.swift
```diff
@@ -0,0 +1,89 @@
...
+        contentView.addSubview(separatorLine)
+
+        setupConstraints()
+    }
+    
+    required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
+
+    private func setupConstraints() {
+        let constraints = [
+
```
DK1DA:
```suggestion
```

Mastodon/Scene/Notification/Notification Filtering/Requests/NotificationRequestsTableViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+                guard let self else { return }
+                self.viewModel.requests = requests
+                var snapshot = NSDiffableDataSourceSnapshot<NotificationRequestsSection, NotificationRequestItem>()
+                snapshot.appendSections([.main])
+                snapshot.appendItems(self.viewModel.requests.compactMap { NotificationRequestItem.item($0) } )
+
+                self.dataSource?.apply(snapshot)
+            }
+        } else {
+            await MainActor.run { [weak self] in
```
DK1DA: You could lift the `await MainActor.run { [weak self] in` out of the `if/else` to only have this once.

Mastodon/Scene/Notification/Notification Filtering/Requests/NotificationRequestsTableViewController.swift
```diff
@@ -0,0 +1,226 @@
...
```
DK1DA: Just the same idea as above, but if you think it's more reasonable to keep the two individual `MainActor.run` calls here then please ignore my comment(s).

Mastodon/Scene/Notification/NotificationTimeline/NotificationTimelineViewModel.swift
```diff
@@ -77,18 +83,45 @@ final class NotificationTimelineViewModel {
                     FileManager.default.cacheNotificationsAll(items: items, for: authContext.mastodonAuthenticationBox)
                 case .mentions:
                     FileManager.default.cacheNotificationsMentions(items: items, for: authContext.mastodonAuthenticationBox)
+                case .fromAccount(_):
+                    //TODO: we don't persist these
```
DK1DA: Maybe rather a`//NOTE:` than a `//TODO:`  ?

MastodonSDK/Sources/MastodonSDK/Query/Query.swift
```diff
@@ -55,6 +55,11 @@ extension PostQuery {
 // PATCH
 protocol PatchQuery: RequestQuery { }
 
+extension PatchQuery {
+    // By default a `PostQuery` does not have query items
```
DK1DA:
```suggestion
    // By default a `PatchQuery` does not have query items
```

MastodonSDK/Sources/MastodonLocalization/Generated/Strings.swift
```diff
@@ -1950,6 +1986,12 @@ public enum L10n {
         }
       }
     }
+    public enum FilteredNotificationBanner {
+      /// Plural format key: "%#@number_of_requests@"
+      public static func subtitle(_ p1: Int) -> String {
+        return L10n.tr("Localizable", "plural.filtered_notification_banner.subtitle", p1, fallback: "Plural format key: \"%#@number_of_requests@\"")
```
DK1DA, line 1992: When using `de_DE` I only see the fallback key here, this looks broken to me...

Bildschirmfoto 2024-07-26 um 10 28 03

kimar (approved):
LGTM 👍  Just a few questions / ideas and one issue with the pluralized format string.

MastodonSDK/Sources/MastodonLocalization/Generated/Strings.swift
zeitschlag: Yeah, I noticed that as well, as you can read [here](https://github.com/mastodon/mastodon-ios/pull/1319/commits/1bf763a94cfdde80f32ea4edb613a78efaeef5a8):

> For whatever reason the plural-translation doesn't have a fallback to Base?!

Once there's a (`de_DE`)-translation, it should be fixed  👍

Mastodon/Protocol/Provider/DataSourceFacade+Notifications.swift
zeitschlag: We don't handle errors elsewhere (i.e. show an alert or something) IIRC, so I don't think so?

Mastodon/Protocol/Provider/DataSourceProvider+NotificationTableViewCellDelegate.swift
zeitschlag: This case isn't supposed to be executed (like `hashtag` before), but due to `DataSourceItem` they just have to be implemented.

zeitschlag: I will change it to `assertionFailure` or `break`, I don't really mind 🙃

PR #1320, opened 2024-06-25
IOS-271: Fix translate functionality leads to empty status content

kimar:
This PR solves the following things:

1. Fixes the post body being set to an empty string in case translation fails
2. Fetches `/api/v1/instance/translation_languages` (only if the `/api/v2/instance` is available) and checks if the transaction between the desired languages is possible
3. Refactors `Instance` usage to now use `MastodonAuthentication.InstanceConfiguration` if feasible with low effort
4. On the home timeline, the reblog status (the wrapper) is taken into account for translation which leads to the translation failing


MastodonSDK/Sources/MastodonCore/MastodonAuthentication.swift
```diff
@@ -113,8 +133,32 @@ public struct MastodonAuthentication: Codable, Hashable, UserIdentifier {
...
-        copy(instanceObjectIdURI: instance.objectID.uriRepresentation())
+    @MainActor
+    func updating(instanceV1 instance: Mastodon.Entity.Instance) -> Self {
+        return copy(instanceConfiguration: .v1(instance))
+    }
+    
+    @MainActor
+    func updating(instanceV2 instance: Mastodon.Entity.V2.Instance) -> Self {
+        guard
+            let instanceConfiguration = self.instanceConfiguration,
```
zeitschlag: Shouldn't a `let instanceConfiguration` be enough (instead of a `let configuration = self....`)?

MastodonSDK/Sources/MastodonCore/Service/InstanceService.swift
```diff
@@ -34,90 +34,64 @@ public final class InstanceService {
             .receive(on: DispatchQueue.main)
             .compactMap { $0.first?.domain }
             .removeDuplicates()     // prevent infinity loop
-            .sink { [weak self] domain in
-                guard let self = self else { return }
-                self.updateInstance(domain: domain)
+            .asyncMap { [weak self] in await self?.updateInstance(domain: $0) }
+            .sink { _ in
+                // no-op
```
zeitschlag: Why is this still needed?

zeitschlag (approved):
Quick testrun was successful, translation works, code looks good.

Good work, thank you! 👍

MastodonSDK/Sources/MastodonCore/Service/InstanceService.swift
DK1DA: Without a `sink` the pipeline is not observed and no events will be emitted.

DK1DA: But we don't need the `_ in` I've removed that 👍

PR #1321, opened 2024-06-26
Show author in Card (IOS-284)

zeitschlag:
This PR updates the Preview Card. Apart from a minor background color change, it introduces the new `StatusCardAuthorControl`, which is visible if the author of an article has a Mastodon-account. If the user taps on the account, the given profile will open.

This feature will be available in Mastodon >4.3 (See [this PR](https://github.com/mastodon/mastodon/pull/30398), for example). HMI, if you need more information on how to test this etc.

Please enjoy these screenshots with some fake data:


Mastodon/Protocol/Provider/DataSourceProvider+StatusTableViewCellDelegate.swift
```diff
@@ -144,6 +144,17 @@ extension StatusTableViewCellDelegate where Self: DataSourceProvider & AuthConte
         }
     }
 
+    func tableViewCell(
+        _ cell: UITableViewCell,
+        statusView: StatusView,
+        cardControl: StatusCardControl,
+        didTapProfile account: Mastodon.Entity.Account
```
DK1DA, line 151: 💡 Just a very minor thing which I think we might do differently in the future: This API has knowledge of the concrete UX (e.g. the user taps because they're on a touchscreen, but when running this on macOS it would probably be described as a "click"), maybe we should name such things rather `didSelectProfile` ?

I know this is currently in line with how it has been done before, just a thought from my side that we might want to change this. WDYT?

kimar (approved):
LGTM 👍  Great work!

PR #1324, opened 2024-07-02
Add authorization to custom emojis

zeitschlag:
This is (mostly) relevant for isolated instances but also these people have a right to custom emojis, right? :D

I encountered this minor issue when working on the Preview-card on an isolated instance.


kimar (approved):
Very good find! :+1:

PR #1325, opened 2024-07-15
Implement Lists and Hashtags Menu (and timeline) (IOS-102)

kimar:
# Implement Lists and Hashtags Viewing Capabilities

This PR adds **List** and **Hashtag** viewing capabilities.

| Menu | Lists | Hashtags |
|---|---|---|
|Simulator Screenshot - iPhone 15 Pro Max - 2024-07-16 at 13 41 07 |Simulator Screenshot - iPhone 15 Pro Max - 2024-07-16 at 13 40 49| Simulator Screenshot - iPhone 15 Pro Max - 2024-07-16 at 13 40 44|


MastodonSDK/Sources/MastodonCore/Service/API/APIService+ListTimeline.swift
```diff
@@ -0,0 +1,26 @@
...
+        let domain = authenticationBox.domain
+        let authorization = authenticationBox.userAuthorization
+        
+        let response = try await Mastodon.API.Timeline.list(
+            session: session,
+            domain: domain,
+            query: query,
+            id: id,
+            authorization: authorization
+        ).singleOutput()
```
zeitschlag: Why do we need this to be a publisher (we don't use)? Is it to be consistent with everything else?

WidgetExtension/Variants/Hashtag/HashtagWidget.swift
```diff
@@ -53,7 +53,7 @@ extension HashtagWidgetProvider {
             do {
                 let mostRecentStatuses = try await WidgetExtension.appContext
                     .apiService
-                    .hashtagTimeline(domain: authBox.domain, limit: 40, hashtag: desiredHashtag, authenticationBox: authBox)
+                    .hashtagTimeline(limit: 40, hashtag: desiredHashtag, authenticationBox: authBox)
```
zeitschlag, line 56: Nice catch!

Mastodon/Scene/HomeTimeline/HomeTimelineViewController.swift
```diff
@@ -140,10 +140,89 @@ final class HomeTimelineViewController: UIViewController, NeedsDependency, Media
...
+                showLocalTimelineAction.state = .off
+                showFollowingAction.state = .off
+            }
+        }
+        
+        let listsSubmenu = UIDeferredMenuElement.uncached { [weak self] callback in
+            guard let self else { return callback([]) }
+            
+            Task { @MainActor in
+                let lists = (try? await Mastodon.API.Lists.getLists(
```
zeitschlag, line 156: If there aren't any lists (or I don't follow any hashtags) but tap on "Lists"/"Followed Hashtags", the submenu disappears right away and this feels kind of broken

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
```diff
@@ -132,6 +132,17 @@ extension HomeTimelineViewModel.LoadLatestState {
                         query: .init(local: true, sinceID: sinceID),
                         authenticationBox: viewModel.authContext.mastodonAuthenticationBox
                     )
+                case let .list(id):
+                    response = try await viewModel.context.apiService.listTimeline(
+                        id: id,
+                        query: .init(sinceID: sinceID),
```
zeitschlag, line 138: More of a gut feeling:

In case of a `timelineContext` switch (from `home` to `public` or `hashtag`, for example), does it make sense to reset `sinceID` explicitly or is it enough just to reset `viewModel.dataController.records = []` ?

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel.swift
```diff
@@ -38,7 +38,11 @@ final class HomeTimelineViewModel: NSObject {
     let isOffline = CurrentValueSubject<Bool, Never>(false)
     var networkErrorCount = CurrentValueSubject<Int, Never>(0)
 
-    var timelineContext: MastodonFeed.Kind.TimelineContext = .home
+    var timelineContext: MastodonFeed.Kind.TimelineContext = .home {
+        didSet {
+            hasNewPosts.send(false)
```
zeitschlag, line 43: Good catch, thanks!

zeitschlag (approved):
LGTM, nice work! 👍 

Just the usual questions and remarks, nothing blocking 🚀

MastodonSDK/Sources/MastodonCore/Service/API/APIService+ListTimeline.swift
DK1DA: Yes this was purely because it's done everywhere like this, but I agree that using streaming Publishers for one-off HTTPS¥ requests doesn't make much sense and it's a good opportunity for refactoring, so I've changed this.

Mastodon/Scene/HomeTimeline/HomeTimelineViewModel+LoadLatestState.swift
DK1DA: We don't need to manually reset it as it will be computed upon entering the new loading state.

Mastodon/Scene/HomeTimeline/HomeTimelineViewController.swift
DK1DA: This makes sense, I looked at how it's done on android and there an empty view is shown, I'll add something similar.

DK1DA: This was the most straight forward way to do:

Bildschirmfoto 2024-07-17 um 13 17 54
Bildschirmfoto 2024-07-17 um 13 17 59

PR #1327, opened 2024-07-20
fix: Remove charset from application/json content type

nikclayton:
Per https://datatracker.ietf.org/doc/html/rfc8259, application/json does not have a charset parameter


kimar (approved):
Thank you! LGTM!

PR #1336, opened 2024-07-31
Fix link char count (IOS-285)

kimar:
This PR fixes the link char count which should be static (by default 23 chars per link) rather than the length of the link itself.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
@@ -343,7 +343,7 @@ extension ComposeContentViewModel {
                     .charactersReservedPerURL ?? MastodonAuthentication.fallbackCharactersReservedPerURL
                 return lengthWithoutLinks + (matches.count * charactersReservedPerURL)
             }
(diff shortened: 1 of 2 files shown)


MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/ComposeContentViewModel.swift
```diff
@@ -324,7 +324,26 @@ extension ComposeContentViewModel {
...
+                    guard let range = Range(match.range, in: input) else {
+                        return 0
+                    }
+                    let url = input[range]
+                    return url.count
+                }).reduce(0, +)
+                let charactersReservedPerURL = authContext.mastodonAuthenticationBox
+                    .authentication
+                    .instanceConfiguration?
+                    .charactersReservedPerURL ?? MastodonAuthentication.fallbackCharactersReservedPerURL
```
renchap, line 343: Do you need the fallback here, if its already checked in `MastodonAuthentication`?

DK1DA: Unfortunately we do because `instanceConfiguration` is optional and might not (yet) exist at this point.

Issue #1381, opened 2024-12-17
[Crash report 💥] Replace Optional Cell Handling with Unwrapped Implementation.


gokulvenkat243:
**Current Code:**
    let cell = tableView.dequeueReusableCell(withIdentifier: "cell"r, for: indexPath) as! TableviewCell

- Here using **as!**. but it’s risky.

**Why Avoid?**

- If the cell is fails, the app will be crashed
- Suitable only when you are 100% confident the cell type is correct

**Instead of:**
guard let cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath) as? TableviewCell  else 
{  
fatalError("Wrong cell") 
}

- This is the safest and most common way to unwrap

PR #1382, opened 2024-12-17
Implement Unwrapped Cells for Improved Type Safety

gokulvenkat243:
Issue: https://github.com/mastodon/mastodon-ios/issues/1381

**Description:**

-    Using Guard let statement they will be prevent the app crashes, when if the cells fails.

### DIFF
--- a/Mastodon/Diffable/Report/ReportSection.swift
+++ b/Mastodon/Diffable/Report/ReportSection.swift
@@ -40,12 +40,18 @@ extension ReportSection {
         return UITableViewDiffableDataSource(tableView: tableView) { tableView, indexPath, item -> UITableViewCell? in
             switch item {
             case .header(let headerContext):
-                guard let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: ReportHeadlineTableViewCell.self), for: indexPath) as? ReportHeadlineTableViewCell else { fatalError("WTF?! Wrong cell.") }
+                guard let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: ReportHeadlineTableViewCell.self), for: indexPath) as? ReportHeadlineTableViewCell else {
+                    assertionFailure("unexpected cell dequeued")
+                    return nil
+                }
                 cell.primaryLabel.text = headerContext.primaryLabelText
                 cell.secondaryLabel.text = headerContext.secondaryLabelText
                 return cell
             case .status(let status):
-                guard let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: ReportStatusTableViewCell.self), for: indexPath) as? ReportStatusTableViewCell else { fatalError("WTF?! Wrong cell.") }
+                guard let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: ReportStatusTableViewCell.self), for: indexPath) as? ReportStatusTableViewCell else {
+                    assertionFailure("unexpected cell dequeued")
+                    return nil
+                }
                 configure(
                     tableView: tableView,
                     cell: cell,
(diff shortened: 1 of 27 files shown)


whattherestimefor:
Thank you for noticing this. Please resubmit with the following change: rather than fatalError(), which will crash in production, return nil when the cell is nil or of an unexpected type:

```
guard let cell = dequeue...(...) as? OurTypeOfCell else {
    assertionFailure("unexpected cell type") // will crash in debug but not in production
    return nil
}
```

Mastodon/Diffable/Report/ReportSection.swift
```diff
@@ -40,12 +40,12 @@ extension ReportSection {
         return UITableViewDiffableDataSource(tableView: tableView) { tableView, indexPath, item -> UITableViewCell? in
             switch item {
             case .header(let headerContext):
-                let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: ReportHeadlineTableViewCell.self), for: indexPath) as! ReportHeadlineTableViewCell
+                guard let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: ReportHeadlineTableViewCell.self), for: indexPath) as? ReportHeadlineTableViewCell else { fatalError("WTF?! Wrong cell.") }
```
whattherestimefor: To avoid crashing in production, please use `assertionFailure("unexpected cell dequeued")` and then return `nil`.

whattherestimefor (changes requested):

Mastodon/Diffable/Report/ReportSection.swift
gokulvenkat243: sure @whattherestimefor

Mastodon/Diffable/Status/StatusSection.swift
```diff
@@ -147,12 +166,15 @@ extension StatusSection {
                 return nil
             case .option(let record):
                 // Fix cell reuse animation issue
-                let cell: PollOptionTableViewCell = {
+                guard let cell: PollOptionTableViewCell = {
                     let _cell = tableView.dequeueReusableCell(withIdentifier: String(describing: PollOptionTableViewCell.self) + "@\(indexPath.row)#\(indexPath.section)") as? PollOptionTableViewCell
                     _cell?.prepareForReuse()
                     return _cell ?? PollOptionTableViewCell()
```
whattherestimefor: This block will never return `nil` (if `_cell` is `nil`, it will create a new `PollOptionTableViewCell()`), so the `else` will never be reached. Best fix would be to remove the `?? PollOptionTableViewCell()`, so that the block _can_ return `nil` and the new `else` can then be reached.

Mastodon/Diffable/Status/StatusSection.swift
```diff
@@ -179,12 +201,15 @@ extension StatusSection {
                 return nil
             case let .history(option):
                 // Fix cell reuse animation issue
-                let cell: PollOptionTableViewCell = {
+                guard let cell: PollOptionTableViewCell = {
                     let _cell = tableView.dequeueReusableCell(withIdentifier: String(describing: PollOptionTableViewCell.self) + "@\(indexPath.row)#\(indexPath.section)") as? PollOptionTableViewCell
                     _cell?.prepareForReuse()
                     return _cell ?? PollOptionTableViewCell()
```
whattherestimefor: same as above

Mastodon/Protocol/Provider/DataSourceProvider+UITableViewDelegate.swift
```diff
@@ -218,6 +221,7 @@ extension UITableViewDelegate where Self: DataSourceProvider & MediaPreviewableV
             parameters.visiblePath = UIBezierPath(roundedRect: mediaView.bounds, cornerRadius: MediaView.cornerRadius)
             return UITargetedPreview(view: mediaView, parameters: parameters)
         } else {
+            assertionFailure("unexpected cell dequeued")
```
whattherestimefor: This is not a dequeued cell being provided to populate the tableview, but rather a cell that is already in the tableview and is being accessed to learn things about its contents. The assertion could be something like "unexpected cell type at \(indexPath)"

whattherestimefor (changes requested):
Thank you for making these changes.  A couple edits to the spots that aren't quite the same as the others and then this should be good to merge.

PR #1387, opened 2024-12-24
 If recent search history is empty show text like No recent searches

gokulvenkat243:
Issue: https://github.com/mastodon/mastodon-ios/issues/1389

Showing No recent searches text, if recent search is empty 

https://github.com/user-attachments/assets/4d09f90f-b4fd-4cb1-9658-9504b9878805


MastodonSDK/Sources/MastodonLocalization/Generated/Strings.swift
whattherestimefor, line 1: Our localization system requires changes be made in a couple files (Localization/app.json and MastodonSDK/Sources/MastodonLocalizations/Resources/Base.lproj/Localizable.strings) and then running swiftgen inside the project directory to update this file.  I will make those changes after merging this.

whattherestimefor (approved):
Thank you!

PR #1392, opened 2025-01-02
support for focus point in image

### LINKED ISSUE #868: [BUG] Image focal point is ignored when displaying attached images

### Current Behavior

When I create a post with the web interface, when I edit an attached image I am able to set a focal point “which will always be in view on all thumbnails.”

This more or less works on the web. (I had set the focal point to be sure the word Restore/Restores was visible.)
Web focus

In the app, the focal point seems to be ignored entirely.

### Expected Behavior

The images shown under the post text will ensure that the circular focal point is fully visible.

### Steps To Reproduce

1. On web, create a post. Attach two images. Set a focal point which is not close to the center.
2. View the post on iPad.
3. Note that the focal point is not visible. It is probably showing the center portion of the images.
iPad not focus

### Environment

```markdown
- Device: iPad Pro 12.9 inch
- OS: iPadOS 16.3
- Version: 1.4.9
- Build: 224
```

### Anything else?

The post I noticed this on was https://mastodon.gamedev.place/@asharp/109649533401033603

tvrrp:
Added support for focus point in image
Size and offset calc are taken from [here](https://github.com/jonom/jquery-focuspoint)

Screenshot 2025-01-02 at 11 38 33

Closes https://github.com/mastodon/mastodon-ios/issues/868

### DIFF
--- a/Mastodon/Protocol/Provider/DataSourceFacade+Media.swift
+++ b/Mastodon/Protocol/Provider/DataSourceFacade+Media.swift
@@ -92,8 +92,13 @@ extension DataSourceFacade {
             
             let mediaView = previewContext.mediaView
 
-            item.initialFrame = {
+            item.initialContainerFrame = {
                 let initialFrame = mediaView.superview!.convert(mediaView.frame, to: nil)
+                assert(initialFrame != .zero)
+                return initialFrame
+            }()
+            item.initialFrame = {
+                let initialFrame = mediaView.contentView().frame
                 assert(initialFrame != .zero)
                 return initialFrame
             }()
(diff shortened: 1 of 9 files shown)


whattherestimefor:
Thanks so much for addressing this. I wonder if this could be done with a little less code by using the anchorPoint property of CGLayer?  Is that an approach you considered or would like to consider?

tvrrp:
@whattherestimefor It can be done using anchorPoint, but in the end, it would require the same amount of code. Much of the code in this MR changes the logic behind the animated transition to MediaPreviewController.

UIImageView clips the original image to fit inside its bounds. Because of this, UIImageView must have the full width or height (depending on the aspect ratio). This allows the origin to be used for shifting the image to the focus point. I think this approach is easier to understand than manipulating layer properties or transforms.

Since I changed the image size, it broke the logic behind the animated transition. That’s why an additional container with clipToBounds enabled was required. Otherwise, the image would pop out of its original container during the push animation, as the code relies on the frame growing to a specific size and assumes that UIImageView will handle fitting the image to the new size.

Changing the anchorPoint of CALayer would result in the same problems as the origin shift approach. The underlying CALayer would pop out of the container during the animation, requiring additional layer rect manipulation in MediaPreviewController.

As for the code behind focus point calculation, it will not work by simply shifting the anchorPoint. The JS library referenced in the [Mastodon documentation](https://docs.joinmastodon.org/api/guidelines/#focal-points) first clips the image to the container using either the container’s width or height, depending on the image’s aspect ratio. From the scaled size, it then calculates the focus point.

whattherestimefor:
Great, thank you!

Issue #1413, opened 2025-02-22
[BUG] handle account switching when accounts have the same id


angristan:
### Is there an existing issue for this?

- [x] I have searched the issues (both open and closed)

### Current Behavior

This is basically like https://github.com/elk-zone/elk/pull/1510.

![Image](https://github.com/user-attachments/assets/f504b864-49d1-4c9d-8091-b34ce53b1295)

In the following example, I have 3 accounts:

| Handle                  | Local ID |
| ----------------------- | -------- |
| `angristan@mstdn.io`    | 1        |
| `angristan@kpop.social` | 1        |
| `admin@mstdn.io`        | 5363     |

-> The ID can't be used to identify an account by itself, we need the instance as well.

I think this could be fixed here: 

https://github.com/mastodon/mastodon-ios/blob/65bf555d9891fe1ac17dbff86f89044934700c70/Mastodon/Scene/Account/AccountListViewModel.swift#L130

By switching to:

```swift
let isActive = activeAuthentication.userID == authentication.userID && activeAuthentication.domain == authentication.domain
```

or something like that. I wasn't able to setup Xcode to be able to run the app locally, so I can't test 😬 

### Expected Behavior

The account switcher should show ✔ only for the current account

### Steps To Reproduce

Connect with accounts on different instances that have the same local ID

### Environment

```markdown
- Device: iPhone 14 Pro
- OS: iOS 18
- Version: v2024.12
- Build:6891
```

angristan:
Ah, looks like `identifer` could work as well

PR #1414, opened 2025-02-22
Fix authentication check to use identifier instead of userID

angristan:
Fix for https://github.com/mastodon/mastodon-ios/issues/1413

| before | after |
| - | - |
| image | image |


whattherestimefor (approved):
Thank you! There is at least one other place in the app that makes a similar mistake. I will make that change shortly. Until then, app behavior may be unpredictable in this situation.

angristan:
thanks! 🙏

Issue #1415, opened 2025-02-24
Mastodon iOS application does not allow users to freely zoom in on GIFs


whattherestimefor:
Opening the gif allows zooming, but it only jumps between regular and one zoomed-in setting.

### Discussed in https://github.com/mastodon/mastodon-ios/discussions/1403

<div type='discussions-op-text'>

<sup>Originally posted by **computerstimulation** February  2, 2025</sup>
There might be only one level of zoom on GIFs on the Mastodon iOS application.</div>

PR #1418, opened 2025-03-02
Fix freely zoom in on GIFs

asynclet:
https://github.com/mastodon/mastodon-ios/issues/1415

## Description of Changes

This pull request adds pinch-to-zoom support for GIF content within the `MediaPreviewVideoViewController`. The functionality is enabled only when the media type is GIF, while the behavior for video content remains unchanged.

### What Has Been Done

- **Added a Container for AVPlayerViewController:**  
  The `AVPlayerViewController` is now embedded inside a dedicated `containerView`. This container isolates the player's view from its internal gesture handling, providing a dedicated target for our custom pinch gesture.

- **Implemented Pinch-to-Zoom Using a Gesture Recognizer:**  
  A `UIPinchGestureRecognizer` is attached to the container view for GIF content. The recognizer updates the container view's transform in response to pinch gestures, allowing users to zoom in and out on GIFs.

Please review the changes and let me know if any adjustments are needed.

### DIFF
--- a/Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
+++ b/Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
@@ -11,13 +11,20 @@ import Combine
 import func AVFoundation.AVMakeRect
 
 final class MediaPreviewVideoViewController: UIViewController {
-
+    
     var disposeBag = Set<AnyCancellable>()
     var viewModel: MediaPreviewVideoViewModel!
     
     let playerViewController = AVPlayerViewController()
-    
     let previewImageView = UIImageView()
+    
+    let doubleTapGestureRecognizer: UITapGestureRecognizer = {
+        let tapGestureRecognizer = UITapGestureRecognizer()
+        tapGestureRecognizer.numberOfTapsRequired = 2
+        return tapGestureRecognizer
+    }()
+    
+    private let containerView = UIView()
     
     deinit {
         viewModel.playbackState = .paused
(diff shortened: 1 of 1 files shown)


Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
```diff
@@ -11,14 +11,21 @@ import Combine
...
-
+    
     var disposeBag = Set<AnyCancellable>()
     var viewModel: MediaPreviewVideoViewModel!
     
     let playerViewController = AVPlayerViewController()
-    
     let previewImageView = UIImageView()
     
+    let doubleTapGestureRecognizer: UITapGestureRecognizer = {
```
whattherestimefor: Does this gesture get used?

Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
```diff
@@ -73,7 +88,25 @@ extension MediaPreviewVideoViewController {
...
+        else { return }
+        
+        gestureView.transform = gestureView.transform.scaledBy(x: gesture.scale, y: gesture.scale)
+        gesture.scale = 1.0
+    }
+}
+
+// MARK: - UIGestureRecognizerDelegate
+extension MediaPreviewVideoViewController: UIGestureRecognizerDelegate {
+    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
```
whattherestimefor, line 101: Which other gestures are in play here that might need simultaneous recognition?

Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
asynclet: No, it’s just some accidental clutter. I’ll remove it.

Mastodon/Scene/MediaPreview/Video/MediaPreviewVideoViewController.swift
asynclet: For default zooming, only the pinch gesture is required. If we want to replicate the behavior of UIImageView, we can use a UIScrollView to contain and pan the content. However, this solution might be overkill.

PR #1433, opened 2025-05-08
Make language variants available in language picker

### LINKED ISSUE #28884: Toots in Chinese regional variants from iOS are not rendered in the correct font
### Steps to reproduce the problem

1. Post a toot from Mastodon on iOS with language set to a specific zh locale code, e.g. `zh-Hant`.
2. Observe the post on Web.

### Mastodon instance

mastodon.social

### Mastodon version

v4.3.0-alpha.0+pr-28693-384ec56
Mastodon for iOS v2023.16 (6043)

### Browser name and version

Safari Version 16.6 (18615.3.12.11.2)

### Operating system

macOS 13.5.2 (22G91)

### Technical details

Inspecting the generated HTML I can see

```html
<div class="status__content__text status__content__text--visible translate" lang="zh"><p>文字</p></div>
```

with `lang="zh"` being the problem there. The `lang` attribute should be marked with `zh-tw` to allow browsers to pick the right font.

## Background

As part of [Han Unification](https://en.wikipedia.org/wiki/Han_unification), Unicode assigns one code point for all regional variants of the same character. Browsers and other text rendering technologies therefore rely on the context (user-specified toot locale, in this example) to pick the right font.

The same Wikipedia article illustrates this:

The resulting text rendering is not unreadable, but it would feel foreign and unprofessional to readers. That being said, I suspect the fix is also relatively trivial (don't even require a new UI) so overall ROI for fixing this issue should pass the bar.

### Expected behaviour

The post is rendered in a font for `zh-Hant`

### Actual behaviour

The post is rendered in a font for `zh`

### Detailed description

(See technical details below)

### Mastodon instance

mastodon.social

### Mastodon version

v4.3.0-alpha.0+pr-28693-384ec56, Mastodon for iOS v2023.16 (6043)

### Browser name and version

Safari Version 17.1.2 (19616.2.9.11.12)

### Operating system

macOS 14.1.2

### Technical details

This the same issue as #27081 but here [the metadata of the example toot](https://mastodon.social/@timdream/111803086845644974) comes with `lang=zh-Hant`, instead of `zh-TW`. Given that `zh-TW` isn't a "valid locale", the fix in #27099 didn't fix it.

(Background: `zh-TW` is a "language + region" locale code, while `zh-Hant` is a "language + script" locale code.)

rschiang:
This PR fixes mastodon/mastodon#28884 by changing how `LanguagePicker` populates its list of available languages.

### What the problem is

Currently, the language picker in Mastodon iOS app builds its list of languages by looping through `Locale.availableIdentifiers`, skipping any regional variants (contains `_` in the locale identifier) and extract only the `languageCode`. This creates several problems:

1) **Languages with multiple script forms overriding each other;**
    - e.g.,`yue-Hans` incorrectly overwrites `yue`, causing Cantonese to be localized in script-specific Simplified Cantonese.
2) **Language families with variants differed enough to be supported [server-side](https://github.com/mastodon/mastodon/blob/main/app/helpers/languages_helper.rb) cannot be honored;**
    - it is impossible to pick `ms-Arab` (Arabic Malay) or `zh-TW` (Taiwanese Mandarin Chinese), for example.
3) **Incorrectly offering 121 unsupported languages;**
    - posts labeled in `ain` (Ainu language) are stripped of their language tags on server-side.
4) **Missing support for 41 languages available [on Mastodon web](https://github.com/mastodon/mastodon/blob/main/app/helpers/languages_helper.rb);**
    - these include Afar, Abkhaz, Avestan, Bislama, Scots, Southern Sami, or Kalmyk.

### What this PR does

To make changes as minimal as possible, this PR only addresses issue (1) and (2). The new iteration algorithm explicitly specifies script and regional locales to be kept and filter out all other non-“base” language variants.

### Other notes

To support (3) or (4) requires us to synchronize language lists across platforms. I’d rather avoid explicitly excluding languages as there might be future support happening on Mastodon Web; it is possible to implement (4) if we polyfill those 41 language locales though (this shall be a separate PR).

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/Language.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/Language.swift
@@ -17,8 +17,8 @@ public struct Language: Identifiable {
     }
     
     init?(id: String) {
-        guard let endonym = Locale(identifier: id).localizedString(forLanguageCode: id),
-              let exonym = Locale.current.localizedString(forLanguageCode: id)
+        guard let endonym = Locale(identifier: id).localizedString(forIdentifier: id),
+              let exonym = Locale.current.localizedString(forIdentifier: id)
         else { return nil }
         self.endonym = endonym
         self.exonym = exonym
--- a/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/LanguagePicker.swift
+++ b/MastodonSDK/Sources/MastodonUI/Scene/ComposeContent/Toolbar/LanguagePicker.swift
@@ -51,15 +51,28 @@ public struct LanguagePicker: View {
     public static func availableLanguages() -> [Language] {
         let locales = Locale.availableIdentifiers.map(Locale.init(identifier:))
         var languages: [String: Language] = [:]
+
+        // these script and variant codes are supported by the server side and are to be kept in list
+        // see SUPPORTED_LOCALES in Mastodon source for more details
+        let variantMappings = [
+            "ms-Arab": "ms-Arab",
+            "zh_CN": "zh-CN",
+            "zh_HK": "zh-HK",
+            "zh_TW": "zh-TW",
+        ]
+
         for locale in locales {
-            if let code = locale.language.languageCode?.identifier,
-               let endonym = locale.localizedString(forLanguageCode: code),
-               let exonym = Locale.current.localizedString(forLanguageCode: code) {
-                // don’t overwrite the “base” language
-                if let lang = languages[code], !(lang.localeId ?? "").contains("_") { continue }
-                languages[code] = Language(endonym: endonym, exonym: exonym, id: code, localeId: locale.identifier)
+            let code = locale.identifier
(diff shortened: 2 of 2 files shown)


rschiang:
Hi @kimar @whattherestimefor, may I request a code review for the PR? :)

This got a [tiny traction](https://g0v.social/@RSChiang/114473211205811255) on the Mandarin fediverse, and I guess the proposed fix do mean something for the users.

rschiang:
@whattherestimefor I’ve re-read the contribution guide and merged the `develop` branch as requested. Could you please take a look and see if anything need to be addressed? Thanks!

whattherestimefor:
@rschiang  Thanks so much for working on this.  To make the changes easier to review, could you please rebase your branch?

In your local checkout, run `git rebase main`
Then `git push --force`

rschiang:
@whattherestimefor @aokj4ck Sorry for taking so long to catch up!

I’ve undone the local merge and rebased the commit—the diff shall be clean now. Please review the code and let me know if anything needs to be improved!

whattherestimefor:
Thank you!

PR #1441, opened 2025-06-02
Implemented hashable properly to adhere to strict equality

RachelRadford21:
This PR is for issue1434 App crashes when user enters "oh' into search bar. The problem was with how the hashable was being implemented. It wasn't checking for strict equality between account id & account. These never change but the url could, which was causing the issue.


whattherestimefor:
Thank you!  I've just made a slightly different commit to fix this before seeing this.  We actually intend to use the user handle + domain as the unique identifier, because each server assigns its own id to each account it is aware of.

PR #1459, opened 2025-09-20
Fix MetricFormatterTests

aokj4ck:
# Description

- `MastodonMetricFormatter` accepts a number such as 1234, and converts it to an abbreviated form: "1.2K" in the US and "1,2K" in the EU.
    - Add MastodonMetricFormatter.abbreviatedGroupingSeparator to allow changing the separator.
    - The default values of abbreviatedGroupingSeparator are: period in the US and comma in the EU.
    - Update the unit tests to explicitly assign abbreviatedGroupingSeparator = "," for EU-styled test assertions.
    - Update the `test_thousandNinetynineFormat()` test function to explicitly test abbreviatedGroupingSeparator with both EU (,) and US (.) separators.
- Fix MastodonTests.testWebFinger() to fulfill the test expectation exactly once, and on the main thread.
- Update MastodonTests.testConnectOnion() to report an XCTFail() instead of an assertion failure so that the process can exit cleanly and report the test failure instead of halting.
    - testConnectOnion is currently failing, there may be additional servers that this works with but I haven't been able to run a passing test.

### DIFF
--- a/MastodonSDK/Sources/MastodonUI/Helper/MastodonMetricFormatter.swift
+++ b/MastodonSDK/Sources/MastodonUI/Helper/MastodonMetricFormatter.swift
@@ -31,12 +31,24 @@ public final class MastodonMetricFormatter: Formatter {
     private let ten_thousands = DecimalUnit.thousand.asInt * 10
     private let ten_millions = DecimalUnit.million.asInt * 10
     
+    ///
+    private let numberFormatter = NumberFormatter()
+    
+    /// MastodonMetricFormatter first converts to decimel _then_ displays the value.
+    /// Do not assign the groupingSeparator.
+    public var abbreviatedGroupingSeparator: String {
+        get {
+            numberFormatter.decimalSeparator
+        }
+        set {
+            numberFormatter.decimalSeparator = newValue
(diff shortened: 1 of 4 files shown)


MastodonSDK/Sources/MastodonUI/Helper/MastodonMetricFormatter.swift
```diff
@@ -25,18 +25,31 @@ enum DecimalUnit: Int {
...
+/// Abbreviate a given number into the highest significant digits and a unit (K for thousands)
 public final class MastodonMetricFormatter: Formatter {
     
     private let ten_thousands = DecimalUnit.thousand.asInt * 10
     private let ten_millions = DecimalUnit.million.asInt * 10
     
+    /// The number formatter instance that will be used. May be customized through ``abbreviatedGroupingSeparator``.
+    private let numberFormatter = NumberFormatter()
+
+    /// MastodonMetricFormatter first converts to decimel _then_ displays the value.
```
whattherestimefor, line 37: typo - decimel vs. decimal

whattherestimefor (approved):

PR #1475, opened 2025-11-02
Run smoke test on multiple device to ensure app can start for sure.

clarmso:


clarmso:
Let me test out the change in my own fork first.

PR #1503, opened 2026-01-16
Align OS deployment target to 18.6+ across all targets

aokj4ck:
- Consolidate iOS Deployment Target configuration in project settings > Project > Mastodon
- Previously each Build Target had their own customized setting for this OS deployment
- Now all Build Targets inherit from the shared Project setting

##### New behavior
Screenshot 2026-01-15 at 23 03 10

##### Previous behavior
Previously the Project had a deployment target of iOS 17, and each Build Target customized the deployment target which duplicates the values and risks misalignment.
Screenshot 2026-01-15 at 23 07 18


aokj4ck:
Previously: #1454, e8b0d89739b3335b2ff0dd5ea9cd21fee70b5074

PR #1508, opened 2026-01-17
Remove some unused classes/symbols

aokj4ck:
# Description

Try out Periphery configuration to scan unused Swift sources

- `brew install periphery && periphery scan`
- Project: https://github.com/peripheryapp/periphery
- Guide: https://adamwulf.me/2024/12/finding-unused-code-with-periphery/

## Limitations

- Periphery can only scan the Mastodon build scheme
- Extension targets (ex: MastodonIntent) fail to build in Periphery
- MastodonSDK Swift Package fails to build in Periphery due to macOS compilation against dependencies
- Due to these limitations there can often be false-positive scan results.
- Periphery is a neat idea but this build system needs more updates to make the most use of it at another time.

## Scan Results

- To avoid spending time on false-positives, this PR aims to focus on removing "**Unused class _____**" results.
- Ignored scan results include unused scoping/accessibility, protocols / protocol conformance, enums, extensions, functions, initializers, parameters, properties, structs, and typealiases.

Deleted files / symbols were reviewed by comparing each modified paths against `git log --format="%h %cI %an %s" develop`. Out of 30 changed files, only 4 were modified in 2025. These four commits cbd048b0e, fa49cd663, bb69eb0e3, deffab82f contain the most-recently changed files that are also modified in this PR; these were reviewed and found to be unused.


aokj4ck:
If this PR is too big I will be happy to split it up.

PR #1514, opened 2026-01-29
feat: improve OAuth compatibility with non-Mastodon servers

vitalyster:
Refactor AccessTokenQuery to use multipart/form-data for OAuth POST requests, enhancing compatibility with servers expecting this format. Additionally, make the `createdAt` field in the Token entity optional to accommodate instances where this information is unavailable.

### DIFF
--- a/MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+OAuth.swift
+++ b/MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+OAuth.swift
@@ -182,7 +182,7 @@ extension Mastodon.API.OAuth {
         
     }
     
-    public struct AccessTokenQuery: Codable, PostQuery {
+    public struct AccessTokenQuery: PostQuery {
         public init(
             clientID: String,
             clientSecret: String,
--- a/MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+OAuth.swift
+++ b/MastodonSDK/Sources/MastodonSDK/API/Mastodon+API+OAuth.swift
@@ -208,19 +208,19 @@ extension Mastodon.API.OAuth {
         public let grantType: String
         
         var contentType: String? {
-            return Self.multipartContentType()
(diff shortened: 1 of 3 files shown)


whattherestimefor:
Thanks for this contribution. We're curious as to which servers require this change? Could you provide an example?

vitalyster:
@whattherestimefor OAuth 2 RFC expects requests with form data, not JSON.
I'm developing Mastodon-compatible server using Spring Security OAuth library and that library does not decode JSON data in OAuth requests. With this patch it works correctly.

vitalyster:
also Mastodon Android client uses form data, not JSON, and works correctly with my server.

grishka:
The Android app does use JSON though. It only ever uses multipart/form-data for file uploads.

vitalyster:
@grishka it may use application/x-www-form-urlencoded, but I'm 100% sure application/json not working with Spring Security OAuth.
I switched json to multipart/form-data because it is already implemented in ios app and it is compatible with mastodon too.

grishka:
I was already 99% certain that it's JSON, but I checked it with a debugger to be extra sure
Снимок экрана 2026-02-09 в 21 37 50

vitalyster:
So, Android app uses wrong content type for OAuth requests too.

vitalyster:
Well, it was easy to switch to URL encoding.
Now, this uses correct `x-www-form-urlencoded` content type, as described in [RFC 6749: The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749), [RFC 8693: OAuth 2.0 Token Exchange](https://www.rfc-editor.org/rfc/rfc8693.html) and [OAuth 2.1 RFC Draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-v2-1/).
Only Mastodon (Ruby/Rails OAuth library?) uses JSON for OAuth requests.
This patch makes iOS app compatible with a lot more OAuth libraries.

PR #1537, opened 2026-03-10
Fix main navigation bar gestures like scroll to top not working on iOS 26

### LINKED ISSUE #1530: [Feature]: Scroll up with Home Button
### Description

The pressure of the Home Button in the tab bar has no effect if the Home Tab is already selected.

### Suggested Solution

To be consistent with the web app, it should scroll all the way up to the most recent post in the timeline.

### Alternatives

_No response_

### Additional Context

Any feedback?

patrys:
Liquid glass takes over the tap gestures and the previously used double tap gesture is never triggered. This change replaces it with handling of tab re-selection inside the `tabBarController(_:didSelect:)` delegate.

Fixes #1530

### DIFF
--- a/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
+++ b/Mastodon/Scene/Root/MainTab/MainTabBarController.swift
@@ -268,12 +268,6 @@ extension MainTabBarController {
         tabBarLongPressGestureRecognizer.delegate = self
         tabBar.addGestureRecognizer(tabBarLongPressGestureRecognizer)
         
-        let tabBarDoubleTapGestureRecognizer = UITapGestureRecognizer()
-        tabBarDoubleTapGestureRecognizer.numberOfTapsRequired = 2
-        tabBarDoubleTapGestureRecognizer.addTarget(self, action: #selector(MainTabBarController.tabBarDoubleTapGestureRecognizerHandler(_:)))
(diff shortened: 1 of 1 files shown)


patrys:
One regression that is not easy to work around is that in `didSelect` it's hard to tell if any views were popped from the view stack. Which means if you scroll down, load a post, and then tap Home, you will not only navigate back out of the post but also scroll to the top, which is annoying (but still better than not having a way to scroll to the top).

patrys:
Actually, scratch the above, the original implementation suffered from the same compromise.

whattherestimefor:
Since the previous solution was implemented with a double tap gesture, how about adding a timing check to this one so that the reselection has to be close in time (probably < 0.5 sec) to the previous one, to avoid too many accidental triggers?

patrys:
@whattherestimefor Did you have a chance to test this on device? I'm not sure it's worth adding a delay since it seems to do some sort of internal debouncing, at least on iOS 26.

We can also later replace the UITabBar with a SwiftUI TabView, which handles this automatically when the tab contains a scroll view.

whattherestimefor:
Yes, we are definitely planning to replace the UITabBar with SwiftUI TabView! (Just not immediately.)

I haven't had a chance to try this on device, yet, will check.

Issue #1402, opened 2025-02-01
[REGRESSION] Notifications pull to refresh never updates contents


whattherestimefor:
### Is there an existing issue for this?

- [x] I have searched the issues (both open and closed)

### Current Behavior

As of version/build:
probably 2025.01

### Expected Behavior

Was last working in version/build: 
2024.12

### Environment

```markdown
- Device:
- OS:
- Version:
- Build:
```

Issue #942, opened 2023-02-15
[BUG] Can't refresh 


rbelew:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I now have to kill the app (toss it away) and restart it to have the updated state of my instance reflected by the app.  

I think this behavior started w/ the v1.4.11 update

### Expected Behavior

Clicking the home button used to create a spinning disc until I got the purple "New Posts" button.  The home button now seems to have no effect, and I can't get the "New Posts" button by down-dragging the timeline, either.

### Environment

```markdown
- Device: i6
- OS: iOS 15.7.3
- Version:v1.4.11
- Build:265
```

Felge-Schneider:
I can confirm this bug on iPhone 13 mini, iOs 16.1.2

samadenbaum:
I have the same issue on iPadOS 16.3.1, but not Android 13

ghost:
My app refreshes only about 1/5 times when randomly opened. iOS 16.3.1

nvsystems:
1.4.12 which was just released has NOT fixed this for me. Refreshing the timeline just does not work.
Opening/closing the app multiple times changes nothing, only force close of the app works.

My guess is that the 1/5 times it works for the previous poster is only when due to memory pressure from using other apps iOS has force closed the mastodon iOS app in the background for him sometimes and then it’s a fresh start.

AInteriorB:
Can confirm this bug. Timeline doesn‘t refresh neither after opening the app nor on down-dragging. Current workaround for me is to change the „tab“ (i.g. go to notifications) and then back to home to force a refresh.

kimar:
Thank you everyone for your input. A fix for this bug will be rolled out shortly.

rbelew:
Well done, works for me now!

PR #1565, opened 2026-08-20
Fix stale notifications after returning from background

### LINKED ISSUE #1554: [BUG] Refreshing notifications not reliable

- [x] I have searched the issues (both open and closed)

### Current Behavior

When going to the notifications menu, new notifications are not displayed until pulling down at least two times to refresh.

### Expected Behavior

Refresh should work after pulling down once or automatic refresh directly after entering the notifications menu.

### Steps To Reproduce

Go to notifications when there should be new ones, the new notifications should not be visible
Pull down once to refresh, nothing happens
Pull down second time to refreh, sometimes it works, sometimes pulling down a third or fourth time is neccessary

### Environment

```markdown
- Device: iPhone 13 Pro, iPhone 17 Pro Max
- OS: iOS 18.6, iOS 26.4.2
- Version: 2026.03
- Build: 7909
```

### Anything else?

_No response_

oliinykdm:
Fixes #1554

The in-app Home and Notifications feeds often stop updating until you force-quit the app. Pull-to-refresh can spin and do nothing; sometimes you have to pull twice. I hit this constantly on my own phone, and the same reports have been coming in for years (#1554, #942, #1402).

The iOS app does not use streaming. Those screens are REST snapshots that are not reliably refetched after the process is backgrounded, and a request frozen by iOS can pin `isFetching` so refresh never runs.

This reloads Home and Notifications when the scene becomes active, abandons stuck fetches, and applies new rows immediately.

Please let me know if I'm wrong or if it is planned to be fixed by an another way. I wish just to make an app feel reliable and not pay for Ivory just because of this.

Thanks!

### DIFF
--- a/Mastodon/In Progress New Layout and Datamodel/FeedLoader/MastodonFeedLoader.swift
+++ b/Mastodon/In Progress New Layout and Datamodel/FeedLoader/MastodonFeedLoader.swift
@@ -88,6 +88,12 @@ public class MastodonFeedLoader<PublishedType: Identifiable, CachedType: Cacheab
         allRecords: [], nextBottomLoad: .initializing)
     @Published private(set) var currentError: Error? = nil
     
+    /// Last time `fetchResults` completed and records were published.
+    private(set) var lastSuccessfulFetchAt: Date? = nil
+    
+    /// Bumped to abandon in-flight fetches that iOS froze in the background or that pull-to-refresh superseded.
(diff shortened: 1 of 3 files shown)


whattherestimefor:
Thank you very much.  We are just reaching the end of a major top-level navigation rearchitecture (currently on a branch). I don't think these changes will apply as-is, but I'll do my best to incorporate them before shipping the next version.

PR #1570, opened 2026-09-12
Rework the Mastodon Shortcut Intents for AppIntents

S0ulf3re:
The Post on Mastodon shortcut action in Shortcuts currently still relies upon the legacy SiriKit Intents framework for posting actions. It would be nice if these intents were updated for the more modern AppIntents system instead.

I'm using Cluade to handle some of this.


S0ulf3re:
I'm not sure if I'm qualified to do this

